PR #43, opened 2014-08-06
Fix 'unreachable code' warnings in latest clojurescript

Skinney:


yogthos:
@Skinney just looking through the prs, looks good to me, could you update it to the latest?

Deraen:
Haven't seen such error so I guess this hasn't been a problem for a while, but I prefer throwing exception to `assert false` so I added that to master: https://github.com/reagent-project/reagent/commit/93043e86c49c62bc04292b735f5bf524aaebb6ba.

yogthos:
👍

PR #84, opened 2014-12-28
add react-with-addons versions to jar

pandeiro:
The react-with-addons files include the useful TestUtils framework React uses, among other things. It would be useful to bundle this up and allow reagent to provide access to that out-of-the-box, at a cost of just two more curl requests. What do you think?


holmsand:
The upcoming version of Reagent will no longer bundle any version of React, instead relying on the new builtin foreign lib handling in ClojureScript. That should also make it much easier to use react-with-addons in the future.

So I'm closing this...

PR #96, opened 2015-02-05
Fix missed reference to `component-function`

mike-thompson-day8:


holmsand:
Good catch! However (I'm happy to say), that was somewhat intentional – I thought I'd make sure that the old API would continue to work for now...

PR #114, opened 2015-03-20
require reagent.ratom macros in cljs namespace

martinklepsch:
Quoting [Clojurescript 0.0-2755 release notes](https://groups.google.com/forum/#!topic/clojurescript/FoiqNV5nunQ):

> The first [improvement] is around macros. If your library does the following:

    (ns foo.bar
       (:require-macros foo.bar))

> Then users that require your library:

    (ns bar.baz
      (:require [foo.bar :as foo]))

> Will get  `foo.bar` macros automatically required and aliased to `foo`.

I think it would be a useful convention to require macros of respective Clojure namespaces. Is there a reason I might be not aware of that this hasn't been done? I'm new to Reagent :-)

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -1,6 +1,7 @@
 (ns reagent.ratom
   (:refer-clojure :exclude [atom])
-  (:require-macros [reagent.debug :refer (dbg log warn dev?)])
+  (:require-macros [reagent.debug :refer (dbg log warn dev?)]
+                   reagent.ratom)
   (:require [reagent.impl.util :as util]))
 
 (declare ^:dynamic *ratom-context*)


Deraen:
:+1:

yogthos:
@holmsand thoughts?

holmsand:
Looks good. I'll take a closer look when I get home this weekend.

martinklepsch:
ping :)

yogthos:
I'm up for merging this in if there's no objections

martinklepsch:
Lets ship this :ship: :smile_cat:

yogthos:
@holmsand @seancorfield @Whoops if there's no objections I'd like to release `0.5.1` with the fix

Whoops:
No objection here.

mike-thompson-day8:
@yogthos  Dan has a few changes in the wings including a fix for #135. https://github.com/holmsand?tab=activity

yogthos:
@holmsand @mike-thompson-day8 fantastic! :)

seancorfield:
As noted in another thread, no need to get a vote from me right now since I'm out of the cljs loop for a while, unfortunately.

danielcompton:
This needs to be added to the CHANGEGLOG.md as well.

PR #126, opened 2015-05-08
Position cursor after an input's value has changed

### LINKED ISSUE #79: Text cursor jumps to last position for `[:input]` fields
The following minimal example

```clojure
(defn my-input [on-change text]
  [:div [:input {:type "text"
                 :value (.toUpperCase text)
                 :on-change #(on-change (-> % .-target .-value))}]])

(def text (reagent/atom "foo"))

(defn main-page []
  [my-input #(reset! text %) @text])
```

will cause the cursor to jump to the last position after each key stroke (try to insert a character in the middle of the string). If I replace `(.toUpperCase text)` with `text` the cursor stays at the correct position.

mike-thompson-day8:
Correctly repositions the cursor - fixes #79

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -91,7 +91,7 @@
   (.! this :cljsInputValue nil))
 
 ;; <input type="??" >
-(def these-inputs-have-a-cursor #{"text" "password" "email" "number" "search" "tel" "url"})
+(def these-inputs-have-a-cursor #{"text" "textarea" "password" "email" "number" "search" "tel" "url"})
 
 (defn has-cursor?
   [input-type]
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -91,11 +91,13 @@
   (.! this :cljsInputValue nil))
 
 ;; <input type="??" >
-(def these-inputs-have-a-cursor #{"text" "textarea" "password" "email" "number" "search" "tel" "url"})
+;; The properites 'selectionStart' and 'selectionEnd' only exist on some inputs
+;; See: https://html.spec.whatwg.org/multipage/forms.html#do-not-apply
+(def these-inputs-have-selection-api #{"text" "textarea" "password" "search" "tel" "url"})
 
-(defn has-cursor?
+(defn has-selection-api?
   [input-type]
-  (contains? these-inputs-have-a-cursor  input-type))
+  (contains? these-inputs-have-selection-api input-type))
 
 (defn input-set-value [this]
   (when-some [value (.' this :cljsInputValue)]
@@ -103,7 +105,7 @@
              (let [node       (.' this getDOMNode)
                    node-value (.' node :value)]
                (when (not= value node-value)
-                 (if-not (has-cursor? (.' node :type))
+                 (if-not (has-selection-api? (.' node :type))
                    ; just set the value, no need to worry about a cursor
                    (.! node :value value)
 
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -90,12 +90,46 @@
 (defn input-unmount [this]
   (.! this :cljsInputValue nil))
 
+;; <input type="??" >
+(def these-inputs-have-a-cursor #{"text" "password" "email" "number" "search" "tel" "url"})
+
+(defn has-cursor?
+  [input-type]
+  (contains? these-inputs-have-a-cursor  input-type))
+
 (defn input-set-value [this]
   (when-some [value (.' this :cljsInputValue)]
-    (.! this :cljsInputDirty false)
-    (let [node (.' this getDOMNode)]
-      (when (not= value (.' node :value))
-        (.! node :value value)))))
+             (.! this :cljsInputDirty false)
+             (let [node       (.' this getDOMNode)
+                   node-value (.' node :value)]
+               (when (not= value node-value)
+                 (if-not (has-cursor? (.' node :type))
+                   ; just set the value, no need to worry about a cursor
+                   (.! node :value value)
+
+                   ;; Setting "value" (below) moves the cursor position to the end which 
+                   ;; gives the user a jarring experience. 
+                   ;; 
+                   ;; But repositioning the cursor within the text, turns out 
+                   ;; to be quite a challenge because changes in the text can be 
+                   ;; triggered by various events like:
+                   ;;    - a validation function rejecting a certain user inputted char
+                   ;;    - the user enters a lower case char, but is transformed to upper.
+                   ;;    - the user selects multiple chars and deletes text
+                   ;;    - the user pastes in multiple chars, and some of them are rejected 
+                   ;;      by a validator.
+                   ;;    - the user selects multiple chars and then types in a single 
+                   ;;      new char to repalce them all. 
+                   ;; Coming up with a sane cursor repositioning strategy hasn't been easy 
+                   ;; ALTHOUGH in the end, it kinda fell out nicely, and it appears to sanely
+                   ;; handle all the cases we could think of.  
+                   ;; So this is just a warning.  The code below is simple enough, but if
+                   ;; you are tempted to change it, be aware of all the scenarios you have handle. 
+                   (let [existing-offset-from-end (- (count node-value) (.' node :selectionStart))
+                         new-cursor-offset        (- (count value) existing-offset-from-end)]
+                     (.! node :value value)
+                     (.! node :selectionStart new-cursor-offset)
+                     (.! node :selectionEnd   new-cursor-offset)))))))
 
 (defn input-handle-change [this on-change e]
   (let [res (on-change e)]


mike-thompson-day8:
In my opinion,  this issue is so important/useful that this single PR would justify an immediate v0.5.1.

marcloyal3:
I have two problems with this on first glance :(

1. It doesnt accomodate custom tags for elements that have a cursor. The set of tags is hardcoded.

2. It doesnt fix why the cursor moves in the first place. It just moves it back again.

Is Reagent by any chance unmounting and remounting the component and retaining only the value prop but not cursor position? If so, i think the solution would be to fix the unnecessary unmounting/remounting

What do you reckon Mike?

mike-thompson-day8:
@marcloyal3  Please be sure to read #79 and the associated links. This patch is a solution for THAT problem.

yogthos:
Would there be a better way to detect whether a tag has a cursor, perhaps using metadata? Otherwise, I think it would at least address the issue in the meantime and if there's a better solution then it could be improved later.

marcloyal3:
I'm concerned about unintended consequence of making Changes to Reagent before investigating if the approach that created the problem is itself problematic. Dan did imply that using non-controlled components would not have the same issue.

mike-thompson-day8:
@marcloyal3  jonase's code pretty clearly shows a problem, I would have thought.  By problem, i mean something jarring and annoying to the user.  Have you run it up to have a look? 

So I don't think there is any doubt about there being a problem. And neither is there any doubt about the cause of the problem.  Again, see the ticket. 

The difficulty has been in finding a solution.  As Dan says in the ticket, the trick to fixing this is knowing how to reposition the cursor.

mike-thompson-day8:
@yogthos the existing check for controlled inputs is here:
https://github.com/reagent-project/reagent/blob/master/src/reagent/impl/template.cljs#L124-L126

yogthos:
Right, if you're filtering for the elements with a particular property then there's really not much else you can do. I'm ok with the merge then unless there are objects.

yogthos:
@holmsand @seancorfield @Whoops are you guys ok merging this in, I think it improves the user experience and could be improved later if needed to address some of the concerns above

Whoops:
I have mixed feeling here. I agree that it's a clear win from a usability perspective if it works as advertised. My main objection would be that the fix probably belongs upstream in React rather than here, but thus far I haven't been able to find anywhere it's been given serious attention on that side.

Overall I'd say I'm :+1:

holmsand:
@yogthos @mike-thompson-day8 This looks very smart, but it's also a bit scary… Browsers are strange when it comes to inputs. 

For example, see: http://stackoverflow.com/questions/21177489/selectionstart-selectionend-on-input-type-number-no-longer-allowed-in-chrome

Perhaps it would be wise to wrap the selection-setting in a try/catch?

I'll take a closer look when I get home this weekend.

antishok:
The only usecase provided (that I could see) which benefits from this, is transforming the text to uppercase/lowercase, but this can (and should?) be done via CSS text-transform.
So, why bother? What's so important/useful about this?

marcloyal3:
@holmsand @yogthos @mike-thompson-day8 @antishok 

There is an easy fix for this that is not scary .... 

Simple add a class to the input component e.g. "upppercase" and in that class specify: 

.uppercase {

text-transform: uppercase;

}

DONE! Works like a charm, and cross browser of course.

marcloyal3:
@antishok  exactly right! Same conclusion reached here.

I've built text editors before that do fancy things and using the text selection API has always landed me in trouble. I can elaborate as soon as ai get a break,

marcloyal3:
@homlsand this seems ti be an issue with React not Reagent and currently React users have to deal with it in the app, so is it Reagent's role to fix React's bugs? Im not sure but i think the better place for this fix would be in React. Take a look here at React Future. Seems like it's on the roadmap.

https://github.com/reactjs/react-future/issues/4

yogthos:
@marcloyal3 I agree that the ideal place to fix this would be in React. However, my feeling is that Reagent users shouldn't have to worry about underlying React quirks and having Reagent provide an intuitive workflow is an important goal. When and if React provides a fix Reagent implementation could be updated to take advantage of it.

marcloyal3:
Let's see what @holmsand comes back with. I'll do some tests with Mike's changes and report back issues if any.

mike-thompson-day8:
@antishok like all good issue creators, Jonase has provided a minimal example to show the problem. That doesn't mean it is the ONLY usecase. 

This patch only affects situations where an <input> has an `on-change` handler for getting char by char changes.  If there's no `on-change` handler (`on-blur` is used instead?), this code never comes into play.

It is for situations where the programmer is using that `on-change` for transform or validation purposes. For example, if the programmer wishes to enforce a certain pattern, like "99.9".  If the user enters an "a" char, then the programmer may wish to reject it immediately.  So `on-change` provides a general transformation and verification mechanism useful in cases where you don't want to wait for `on-blur` or form submission. 

@holmsand thanks for that link.  I will take "number" and "email" out of the list of input types with cursors and add that link into the code. 

@marcloyal3 I'm puzzled. That link you give seem to prove the opposite of what you say. From the state of that issue (created a year ago) and that repo, it seems to clearly NOT be on any roadmap.

yogthos:
It sounds to me that if there aren't any regression this change improves the state of things overall, so my vote would be to merge this in.

antishok:
@mike-thompson-day8 apologies, I had tested the validation usecase before, and thought I saw it working fine without your patch, but testing again now I see I was wrong.

Whoops:
Hmnm.. I'm not sure the solution is to remove `number` and `e-mail` from the list, since it still would work as expected in other browsers, and we may still want to use this kind of transformation on them. It also doesn't deal with the case that it gets removed from say `tel` or `password` at a later date. I think it makes more sense to deal with the error there and regress to the current behavior when we can't set `selectionStart/End` as @holmsand suggested,

mike-thompson-day8:
@Whoops the link provided by that stackoverflow answer was to:
https://html.spec.whatwg.org/multipage/forms.html#do-not-apply
This spec says when/where `selectionStart` and `selectionEnd` should exist. I was working off that spec when I removed "number" and "email", and you'll see I've now included that link in the patch as a reference.
But, sure, if the spec is not well implemented and not to be believed, then putting in a try catch is trivial.

marcloyal3:
@mike

They ptobably just dont know what yo do about it, nor that they don't know about it or dont wish to fix it. 

Would you be able to patch React instead of Reagent? Just curious. At least then we'd attack the problem at the source. And you would be making a contribution to the mothership technology

marcloyal3:
@holmsand several posts i found attribute the reason of cursor jumping to async setState. There is a really good explanation here:

http://stackoverflow.com/questions/28922275/in-reactjs-why-does-setstate-behave-differently-when-called-synchronously

marcloyal3:
@all 

Here's what's happening.

Synchronous setState

you press X
input.value is 'HelXlo'
you call setState({value: 'HelXlo'})
the virtual dom says the input value should be 'HelXlo'
input.value is 'HelXlo'
no action taken

Asynchronous setState

you press X
input.value is 'HelXlo'
you do nothing
the virtual DOM says the input value should be 'Hello'
react makes input.value 'Hello'.
Later on...

you setState({value: 'HelXlo'})
the virtual DOM says the input value should be 'HelXlo'
react makes input.value 'HelXlo'
the browser jumps the cursor to the end (it's a side effect of setting .value)

marcloyal3:
Copied & pasted from S.O.

mike-thompson-day8:
@marcloyal3   The story so far:
  1. Jonase has already done an excellent investigation and written it up: http://jonase.github.io/nil-recur/posts/14-12-2014-controlled-input.html
  2. Dan has already commented on that investigation and explained what's hard about a solution.  
  3. This patch is an attempt to solve what Dan identified as difficult.

If you want to come to terms with the problem (stage 1 above) can I suggest that you read #79 carefully.  Honestly, I think you'll find everything there already.

And, I'm not sure if you were serious or not, but no, I'm not patching React.  But thanks for the suggestion :-)

holmsand:
Also, patching React probably won't help. Reagent has to do its own handling of inputs, since React doesn't like async updates.

marcloyal3:
@holmsand if React cannot be patched to solve the issue with async update of inputs then it has to be solved in Reagent. 

When you say "React doesnt like async updates" what does it mean as far as this problem and its solution. Can you educate us a bit? 

Thank you!

marcloyal3:
@mike

Great article and for sure explains the problem in its clearest form.

But its conclusion is startling: Om/Reagent's use of rAF for batching updates is essemtially a do-it-at-your-own-risk strategy then are we able to make an exception in case of inputs and update synchronously or is the only thing we can do is recall the cursor selection state and apply it after value is reset? 

So very curious

Also, i am sure you can make contributions to React that would affect Reagent positively guven  your deeper than average understanding of Reagent and React :-)

JulianLeviston:
I just encountered this. It seems to happen only when I type fast in Safari. I'm not doing any capitalisation or anything at all to the input, but rather just have a controlled textarea that has a fair amount of processing driven off it (now using a go block). I'm about to switch to using a debouncing core.async channel for the "processing driven off it", which I'm guessing will stop this being an issue in my case, but it's interesting, and I can't release my software like this. (Planning on a release within the next two days).

The weird thing is, I didn't have this issue up until I recently changed some more code to run async. I've been trying to stop the text input to the UI from lagging badly (due to the processing I had driven off it), so I've been wrapping some things that were blocking the UI to delay execution, and replacing others by using add-watch on ratoms.

I feel like what is probably happening is the async code is causing a delay in reagent updating its atoms where React's VDOM is actually jumping in and rendering the previous value over the top of the keystroke, which, when replaced is causing the cursor jump to occur. Kinda sucks. Pity there's not a way to set the value initially, then not sync, like with a non-live cursor deref or something (deref-once atom-name) rather than (deref atom-name). But this wouldn't solve the other issues people seem to be having.

yogthos:
@holmsand @seancorfield @Whoops 

I'm inclined to merge this in, it improves user experience and I can't see any obvious downsides to this. If a better approach is discovered later this can always be updated to use it.

Whoops:
:+1: here. When you mentioned a 0.5.1 I actually immediately thought of this PR.

JulianLeviston:
Does this work with text-replacement? It seem kind of hacky.

yogthos:
@JulianLeviston any chance you could test that :)

mike-thompson-day8:
@JulianLeviston We tested with various text-replacement scenarios and it worked correctly.   By "text-replacement", I assume you mean:  (1) user selects some text  (2) user then types and, in so doing, "replaces" the selected text.  

We tested cases where the initial "selection" included:
   - all the text
   - the first char of the text
   - the first character plus some others
   - the last character of the text
   - the last character plus some others 
   - the middle of the text  (not including first of last chars)

We also tested "selecting text" and then hitting delete.  We tested "selecting text" and hitting backspace.  Etc. Those cases worked also.

Bottom line is this:  even if there is some edge case that we somehow haven't covered, this patch remains an improvement on what there is now because what's there now never works, ever.

seancorfield:
I haven't dived in when my name was called because, unfortunately, we abandoned our ClojureScript work (possibly just for a while, possibly for a long time) and so I don't have much of an opinion right now. I do vaguely recall seeing a problem with typing into input fields occasionally but it wasn't common and it didn't seem reproducible (now I know why!). If @mike-thompson-day8 says this improves the UX and has been well-tested, I see no reason to object. After all, if it is found to cause other problems, it could always be reverted and we'd be no worse off than when we started :)

JulianLeviston:
@mike-thompson-day8 First up, you guys are *much* better at this stuff than me, so don't take these particular observations with much other than a pinch of salt. The thing that seems a bit "off" is it feels like we're fighting the browser here. If we have a capitalisation algorithm, it should CANCEL the input event, then MUTATE the contents of the text field, using diffing, just like react does with everything else. I'm not actually 100% sure why it doesn't already do this, but I guess it's too fine-grained. So, perhaps this is a React issue, and like you say, if it's better than what's there, this should be applied as a temporary bandaid.

Feel free to just ignore this if it's not important. I'll understand. And, apologies for the noise if it is.

mike-thompson-day8:
@JulianLeviston If you are interested, have a look through #79 and the discussion above.

JulianLeviston:
@mike-thompson-day8 yeah, thanks. I've been following that discussion, too. The issue isn't whether capitalisation is a text-transform and therefore belongs as presentation, though... the issue applies to all manner of things such as filtering or other non-presentation-based transformations. I guess we just don't have enough control over the browser in this instance. I simply don't know enough about how the DOM represents text field value contents to know whether they're mutatable, and the impact this has on the selection. You're right, though, that this patch would solve 80% of the issues people have.

JulianLeviston:
@mike-thompson-day8 Just FYI, the following project is what happens when you proceed down this path a long way... (not sure if you're familiar with this project - might be worthwhile looking at)... 

https://github.com/bitovi/syn

It's very interesting.

Anyway, I'll be quiet now ;-) Thanks!

idibidiart:
@JulianLeviston 

Hahaha re: the Syn project! First bit of humor I hear in this discussion.

mike-thompson-day8:
@holmsand   bump

joodie:
Hello good people,

I'm running into the issue described simply by running the on-change events through a core.async channel and updating the app state atom in a go-loop from the events in the channel. Any typing in an input field moves the cursor to the end of the input field. This makes my application pretty much unusable and working around it is going to be painful.

I've put a standalone demonstration of the problem at https://github.com/joodie/reagent-issue-79-demo/blob/master/src/cljs/reagent_issue79_demo/core.cljs 

This patch completely fixes the issue for me, so I would like to know if it's going to be merged. If there's anything specific blocking the patch and/or if I can anything do to speed up the process or fix the issue in some other way, please let me know!

Joost.

Whoops:
@yogthos @mike-thompson-day8 I think we should go ahead and merge this in. I know we've kind of looked for @holmsand's input, but it's not looking like we're going to get it anytime soon and he seemed tentatively supportive previously. Any objections?

yogthos:
I say let's do it :)

Whoops:
I think we know @mike-thompson-day8's vote, so BAM! Done. Do we have a release coming up or should we maybe cut one for this change?

yogthos:
Now we just have to cut a release, @holmsand  or @mike-thompson-day8 would one of you be able to do that?

Whoops:
@yogthos  I can tomorrow if no one has beat me to it, I'm off to dinner so it's not likely to get done this evening. You and I both have access to cut a release but @mike-thompson-day8 doesn't at the moment. You should be able to grant him access at https://clojars.org/groups/reagent, however if you find out his name on Clojars.

yogthos:
@Whoops awesome! :+1:  I can add @mike-thompson-day8 on Clojars as well for the future. :)

mike-thompson-day8:
I'm on holidays ATM so limited acces or time. Back next week

holmsand:
Looks good! I'll upload a new release tomorrow (or possibly a quick rc, to get some testing for peace of mind...).

joodie:
Great! :+1:

holmsand:
There is now a 0.5.1-rc that could use some testing :)

yogthos:
:+1:

Whoops:
Great, I'll upgrade our apps and see if there are any issues.

yogthos:
already updated a few here in dev and so far no problems :)

aft-luke:
Given the implementation removed `email` and `number`, what are we supposed to do with those cases?

JulianLeviston:
@aft-luke Not seeing where this removes anything to do with `email` or `number`. Could you help me understand that a bit better?

aft-luke:
Sorry for the vagueness.

> This spec says when/where selectionStart and selectionEnd should exist. I was working off that spec when I removed "number" and "email", and you'll see I've now included that link in the patch as a reference.

The patch to fix the jumping caret previously had number and email, but now does not because browser specifications say it shouldn't. Therefore the merged patch does not include "email" and "number" types in the jumping caret fix. Those two input types, then, still experience a bad caret experience.

JulianLeviston:
@aft-luke oh bravo! thank you that's *much* clearer now. I suppose this is browser-centric, right? Meaning mobile safari won't use a text control to do input on `number`, for example, but desktop safari, chrome & ff *will*?

src/reagent/impl/template.cljs
```diff
@@ -90,12 +90,48 @@
 (defn input-unmount [this]
   (.! this :cljsInputValue nil))
 
+;; <input type="??" >
+;; The properites 'selectionStart' and 'selectionEnd' only exist on some inputs
+;; See: https://html.spec.whatwg.org/multipage/forms.html#do-not-apply
+(def these-inputs-have-selection-api #{"text" "textarea" "password" "search" "tel" "url"})
```
GetContented: maybe `inputs-with-selection-api` might be a bit better.

PR #141, opened 2015-06-05
closes #135; not clear what dirty? was intended to do

### LINKED ISSUE #135: Strange bug with cursor and :on-blur
Hi,

Sorry about ambiguous title but the problem is so strange that I can't describe it better.

We've had strange problems in our project where sometimes Reagent stops responding to state changes after changing input's value. Yesterday we found that this is caused by a ` :on-blur` handler which updates the state through a cursor.

I have a demo here: http://deraen.github.io/problems/reagent-strange-bug/
and the source can be found here: https://github.com/Deraen/problems/blob/reagent-no-render/src/hello_world/core.cljs#L11-L29

1. app-state is saved on single ratom
2. forms access (:value, :on-change, :on-blur) the data through a cursor
3. actions (save thingie etc.) access state atom directly
4. `:on-blur` does something like `#(swap! data assoc :not-pristine true)`

To reproduce:

1. Change the input value
2. Click save => the component is re-rendered successfully
3. Change the input value
4. Click save => the component is not re-rendered
5. Changing the input value now will cause component to be re-rendered

Everything works if

- Cursor is not used
- A non-empty map is assoced in `:on-blur` handler instead of empty map or true

savaki:
This appears to fix the issue.  While the tests all pass, I wasn't able to come up with a test case to reproduce this.  Could someone chime in as to whether dirty? is needed?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
+# ignore intellij files
+.idea
+*.iml
+*.ipr
+*.iws
+
 target
 pom.xml
 .lein-repl-history
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -240,7 +240,7 @@
 
   IComputedImpl
   (-handle-change [this sender oldval newval]
-    (when (and active? (not dirty?) (not (identical? oldval newval)))
+    (when (and active? (not (identical? oldval newval)))
       (set! dirty? true)
       ((or auto-run run) this)))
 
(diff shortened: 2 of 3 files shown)


yogthos:
@holmsand is it ok to merge in?

Deraen:
I'm thinking this can be closed as cdd3c85 was identical change.

yogthos:
looks like it

PR #144, opened 2015-06-09
Make use of deftype constructor instead of interop

EduardoMRB:
As Stuart Sierra stated in his [blog post](http://stuartsierra.com/2015/05/17/clojure-record-constructors)
we should use `deftype` and `defrecord` constructor functions or create
our own constructor functions for types and records.

### DIFF
--- a/src/reagent/impl/batching.cljs
+++ b/src/reagent/impl/batching.cljs
@@ -63,7 +63,7 @@
       (run-queue q)
       (run-funs aq))))
 
-(def render-queue (RenderQueue. (array) false (array)))
+(def render-queue (->RenderQueue (array) false (array)))
 
 (defn flush []
   (.run-queue render-queue))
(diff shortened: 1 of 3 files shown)


JulianLeviston:
It'd be nice if there were reasons given for doing things. "Because Stuart Says So" isn't a true reason.

In this case, it's pretty obvious that he's advocating not relying on the current implementation of Clojure because it may change in the future, but cargo culting is just something to watch out for.

EduardoMRB:
Hi @JulianLeviston ,

I think I gave the wrong impression, sorry about that.

I believe that using constructor functions make the code more clear. If I see `->Foo`, I know that `Foo` is either a record or a type and not a java or js class.

JulianLeviston:
Fair point! :)

mike-thompson-day8:
I think this should be merged (although currently has conflicts).

EduardoMRB:
@mike-thompson-day8 I think it's ok now.

Deraen:
Merged in https://github.com/reagent-project/reagent/pull/288

PR #149, opened 2015-06-25
142 extended hiccup

hlship:
I've done what I can, you can see I added another issue about how to run tests & etc.

I've tested this in a branch of my project and it works fine as far as I can tell.  It's not something that you use everywhere, but it can be nice (especially with things like Bootstrap, which can deeply nest elements).

If this gets accepted, I have ideas about making it easier to include XML/HTML entities. I sure would like an easy way to &nbsp; , etc.

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -227,7 +227,7 @@
             (make-element argv comp p first-child)))))))
 
 (defn- expand-tags
-  "Used to supported the extended Hiccup syntax for nested elements. In addition to the keyword
+  "Used to support the extended Hiccup syntax for nested elements. In addition to the keyword
   specifying tag, optional id, and optional class(es), the '>' character indicates a nested element.
 
   e.g.
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -261,7 +261,7 @@
                (assert (valid-tag? tag)
                        (str "Invalid Hiccup form: "
                             (pr-str v) (comp/comp-name)))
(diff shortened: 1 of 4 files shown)


src/reagent/impl/template.cljs
```diff
@@ -226,17 +226,51 @@
...
-                 (pr-str v) (comp/comp-name)))
-    (if-some [ne (native-element tag v)]
-      ne
-      (reag-element tag v))))
+  (let [tag  (nth v 0)
+        tags (do
+               (assert (valid-tag? tag)
+                       (str "Invalid Hiccup form: "
+                            (pr-str v) (comp/comp-name)))
+               (if (keyword? tag)
```
seancorfield: It seems that if `tag` is not a keyword, then `tags` will be `nil` -- is that intentional? If so, `when` would be better than `if` here.

seancorfield:
It looks like a really nice enhancement -- and a fairly small, localized change! Very nice! I posted a question on one line of the diff.

src/reagent/impl/template.cljs
hlship: Agreed. I'll update the PR.

hlship:
Thanks. Also, I believe that use of this extended syntax, when appropriate, will generate smaller compiled JavaScript. The compiled JavaScript for creating a nested collection of vectors is larger than the syntax for a single keyword. That leaves a concern about performance, but I don't have the tools to identify if that's a problem (I highly doubt it).

pkobrien:
I agree with Sean. That looks really handy. Nice!

Issue #125, opened 2015-05-05
clojure 1.7.0 warning


mike-thompson-day8:
Using this dependency:
    `[org.clojure/clojure "1.7.0-beta2"]`

this compiler warning is produced:  
```
run! already refers to: #'clojure.core/run! in namespace: reagent.ratom, being replaced by: #'reagent.ratom/run!`
```

mike-thompson-day8:
Just to save on a few google searches ...  

It turns out that clojure 1.7.0 has a new function `clojure.core/run!`  
https://github.com/clojure/clojure/blob/clojure-1.7.0-beta2/src/clj/clojure/core.clj#L7363-L7368

seancorfield:
FYI, `update` and `cat` are also new in 1.7.0 and those caused a lot of library warnings as well.

JulianLeviston:
Odd. I wonder why I don't get those warnings. Is this actually a namespacing issue, in which case, can it be closed?

mike-thompson-day8:
To see it, you have to be requiring the macros in `reagent.ratom`

JulianLeviston:
Ok thanks. Please excuse the noise then! :)

yayitswei:
I'm seeing this warning as well. `run!` is an unfortunate name for a core function because quite a few libraries have a run! function.

Frozenlock:
There's `:refer-clojure` for this purpose:
https://clojuredocs.org/clojure.core/refer-clojure

skrat:
Quick grep reveals that this macro is only used in test suite. Why not moving there then?

mike-thompson-day8:
It has has been used by many. Just like `reaction`

yogthos:
@mike-thompson-day8 I'm not seeing the warning with the latest, is this still a problem?

danielcompton:
(On behalf of Mike) nope, looks like it was fixed in https://github.com/reagent-project/reagent/commit/569322949b31921a5eeedb4158414da86027ea49.

PR #150, opened 2015-06-25
rename run! macro to react!

darwin:
under [org.clojure/clojurescript "0.0-3308"]
it was shadowing clojure.core/run! and causing warning:

```
WARNING: run! already refers to: #'clojure.core/run! in namespace: reagent.ratom, being replaced by: #'reagent.ratom/run!
```

### DIFF
--- a/src/reagent/ratom.clj
+++ b/src/reagent/ratom.clj
@@ -5,7 +5,7 @@
   `(reagent.ratom/make-reaction
     (fn [] ~@body)))
 
-(defmacro run!
+(defmacro react!
   "Runs body immediately, and runs again whenever atoms deferenced in the body change. Body should side effect."
   [& body]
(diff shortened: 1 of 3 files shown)


mike-thompson-day8:
See also  #125

darwin:
Thanks for the link.

AFAIK the problem with `:refer-clojure` is that it would have to be written in every file including ratom macros.

Also I think that run! name is unfortunate. It implies something which is run once and has side-effects. We need some better name which hints "this piece of code is being run multiple times in reaction to ratom changes".

danielcompton:
If we're going to go down this track, it would be good to go through a depreciation process for this. `run!` should still be usable for at least one more minor version series to let people update their code gracefully, and deprecation warnings should be printed when people use `run!`.

Deraen:
Closing but documented in #615

PR #158, opened 2015-07-23
Revert #128 and add tests for ISwap implementations

moskvax:
I'm not sure what the motivation behind #128 was, as the & syntax doesn't work with deftype and the core swap! function does this destructuring already. I added some tests for the old ISwap implementations, and they seem to work as expected.


Deraen:
`-swap!` vararg arity was fixed in https://github.com/reagent-project/reagent/commit/4f629d0a34cf741387cabd4636a36ecacaf9a9dc but current tests do not test the arity, so additional tests are a good change.

Issue #161, opened 2015-07-28
Renaming atom to ratom


danielcompton:
EDIT: This issue was opened to address the common issue of people using Atoms instead of Ratoms. Initially it was to investigate the possibility of logging a warning when users rendered a Clojure atom instead of a Ratom. That was deemed infeasible, but the much easier renaming of atom to ratom came up as another avenue. I'm editing this issue rather than creating a new one as there was a lot of good discussion around renaming atom to ratom.

***

ORIGINAL: Is it worth logging a warning to the user when they use an atom instead of a ratom in Reagent rendering code? There's arguments to both sides, so I'm opening up an issue for discussion.

Frozenlock:
I'm still using vanilla atoms with Reagent.

Do you have a way of distinguishing an accidental atom versus a legitimate one?
What would be the overhead?

danielcompton:
I'd check if it implemented `IReactiveAtom` I think.

Frozenlock:
But how would that find accidental atoms? Like I said, I still use vanilla atoms when I don't need it to update a component.

danielcompton:
It depends what you mean by accidental. What's the benefit of using Clojure atoms over just using ratoms everywhere?

Frozenlock:
Less overhead. Better control on what will trigger an update.

There might also be cases where one has no choice. 
Other atom implementations can exist, and Reagent does not allow them to be reagent-ized. (See #92)

Assuming everyone will always use ratoms is... well... a huge assumption. :smile:

JulianLeviston:
@Frozenlock I agree. @danielcompton I don't like this idea much. It feels too magic, and takes control away from the programmer.

danielcompton:
A few things:
1. After talking with colleagues, I'm not sure how this could actually be implemented.
2. We're only talking about logging a warning, I'm not sure what's magical about that?
3. There would probably need to be a flag you could set that says "I know what I'm doing, don't warn me".
4. Another route to take would be to look at renaming atom to ratom, so people could never use one instead of the other accidentally. This would need to go through a deprecation process of course.

mike-thompson-day8:
There's a **very** steady stream of people who get caught by this, and it is a baffling problem to solve.

But equally there is absolutely no way to solve this technically.  There is no way to know if an `clojure.core/atom` is being referenced where a `reagent.core/atom` is intended.  

[...]

PR #164, opened 2015-07-31
Make references to atom explicit, not referred

JulianLeviston:
After extensive discussion in https://github.com/reagent-project/reagent/issues/161 it appears a number of new people get confused by the use of the referred reagent atom (because it's named the same as clojure.core/atom). Changing this page (and other parts of the documentation) to use an aliased namespace directly without using a referred version of atom *may* decrease the number of confused people. This is largely an experiment to see if it helps. (It can't hurt!) :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,13 +85,13 @@ assuming we have imported Reagent like this:
 
 ```clj
 (ns example
-  (:require [reagent.core :as reagent :refer [atom]]))
+  (:require [reagent.core :as r]))
 ```
 
 State is handled using Reagent's version of `atom`, like this:
 
 ```clj
-(def click-count (atom 0))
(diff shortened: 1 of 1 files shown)


danielcompton:
This page also does it: http://reagent-project.github.io

JulianLeviston:
@danielcompton yes, the whole project there does it... there are in total about 15 references, I think... I just wanted to test the water to make sure this was something that @holmsand wanted to enact before going ahead with mass change and pull requests ;-)

holmsand:
@JulianLeviston Looks good to me. Thanks! I've already started to do the same in the rest of the documentation and examples.

JulianLeviston:
<3

PR #167, opened 2015-08-02
Slight improvement to make-reaction

GetContented:
Just changes a call to (not (nil? derefed)) to the already calculated active val.

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -323,7 +323,7 @@
         reaction (Reaction. f nil dirty active
                             nil nil
                             runner on-set on-dispose)]
-    (when-not (nil? derefed)
+    (when active
       (when debug (swap! -running inc))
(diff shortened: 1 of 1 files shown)


mike-thompson-day8:
Looks okay to merge.  @holmsand ?

holmsand:
@mike-thompson-day8 Sure – but the patch will become obsolete very soon (I just managed rid of `active` in a local branch...).

Btw: Big thanks for all the issue handling!

GetContented:
@mike-thompson-day8 there are a bunch of 2 year old stale branches that might need cleaning, too.

@holmsand this isn't "active?" it's a local var called "active", by the way.

holmsand:
@GetContented Yes, both `active?` and `active` are going away...

GetContented:
@holmsand sweet :) no need to merge then. :)

PR #172, opened 2015-08-07
Change r/atom defs to defonce in readme

GetContented:
For better compatiblity with figwheel, change the two references that are def'd with r/atom values to be defonce so they persist through code updates.


Whoops:
Thanks!

PR #178, opened 2015-08-21
Support component getChildContext and childContextTypes

sundbry:
This adds basic support for the experimental "Context" feature of react.js.

https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html

It adds two properties to reagent.core/create-class:

:child-context-types {:key type}, a dictionary specifying the types of objects returnable by the component's context,
:get-child-context (fn [this] ...) a function returning the context of a reagent component

### DIFF
--- a/src/reagent/core.cljs
+++ b/src/reagent/core.cljs
@@ -109,14 +109,16 @@ Returns the mounted component instance."
   "Create a component, React style. Should be called with a map,
 looking like this:
 {:get-initial-state (fn [this])
-:component-will-receive-props (fn [this new-argv])
-:should-component-update (fn [this old-argv new-argv])
-:component-will-mount (fn [this])
-:component-did-mount (fn [this])
-:component-will-update (fn [this new-argv])
-:component-did-update (fn [this old-argv])
-:component-will-unmount (fn [this])
-:reagent-render (fn [args....])   ;; or :render (fn [this])
+ :get-child-context (fn [this])
+ :child-context-types {:key type}
+ :component-will-receive-props (fn [this new-argv])
+ :should-component-update (fn [this old-argv new-argv])
+ :component-will-mount (fn [this])
+ :component-did-mount (fn [this])
+ :component-will-update (fn [this new-argv])
+ :component-did-update (fn [this old-argv])
+ :component-will-unmount (fn [this])
+ :reagent-render (fn [args....])   ;; or :render (fn [this])
 }
 
 Everything is optional, except either :reagent-render or :render.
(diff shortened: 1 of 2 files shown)


mike-thompson-day8:
I don't think this treatment is complete.  

It correctly adds `getChildContext` and `childContextTypes` but misses `contextTypes`

seantempesta:
Would love to have this.  I'm currently doing this, which seems a little hacky:

```clojure
(defn main [{:keys [navigation]}]
  (this-as this
    (let [showActionSheetWithOptions (aget this "_reactInternalInstance" "_context" "showActionSheetWithOptions")]
...
```

sundbry:
>  I'm not convinced this would be very useful with Reagent. I believe we have better ways...

@Deraen This misses the point. The context is added for the benefit of other libraries/components, which may require something in the React context.

mike-thompson-day8:
It would be  useful to `re-frame` to have this feature or, preferably, a more CLJS friendly version of it.

The nub of the idea is that a Component can search "up" through the "parent" hierarchy for values.  (is it "parent" or "owner", I can never remember the difference?)  Child components can then discover "context" information, making it easier to create reusable components.

Just to be clear: in theory this feature is unneeded because values could have been passed down from parent to child, all the way down. So it is a convenience but a mighty useful one, which could bring a lot of simplicity if done correctly. 

It is also a feature used by Redux, so currently about 70% of the React world depends on it.  And that means its "experimental" status is misleading IMO. I can't see how it could ever be removed at this point. 

But the pure React "context" approach involves as lot of boilerplate which seems annoying when using ClojureScript, so I've long wondered about an alternative, more ClojureScript friendly approach to the same thing. 

Any ideas?

Deraen:
> The context is added for the benefit of other libraries/components, which may require something in the React context.

Good point. I was thinking context would be more like something that would be used by apps, not libraries.

> But the pure React "context" approach involves as lot of boilerplate which seems annoying

Always having to add `contextTypes` to access context does indeed seem annoying. Perhaps we could create something like `reagentContext` which would be Clojure datastructure, and would be accessible without adding `contextType` (we could always add this to contextTypes maybe?

Deraen:
Hi,

I have added test to Reagent for using component context: https://github.com/reagent-project/reagent/commit/a8ec0d219bbd507f51a4d9276c4a1dcc020245af

This didn't require any changes to Reagent implementation, as long as user takes care of returning JS objects, and knows that when using additional lifecycle methods (like `get-child-context`) `this` is not available as first parameter, but can be accessed using `this`.

I think this should take care of using any JS libraries using context?

I propose leave the "Cljs friendly" version of this for later. In any case, we might want to keep JS and Cljs versions separate, to make sure Cljs implementation doesn't break some JS use cases.

PR #186, opened 2015-09-12
Handle sequences in hiccup like hiccup does

timothypratley:
Splices in sequences so you don't need to write syntax to do it:

```clojure
[:div
   [:h2 "Seq of divs without ids"]
   (for [i (range 3)]
     [:div (str "A" i)])

   [:h2 "Nested sequences"]
   (for [i (range 3)]
     [:div
      (for [j (range 3)]
        [:div (str "G" i j)])])

   [:h2 "Sequence of divs"]
   '([:div "D0"] [:div "D1"])
   [[:div "D2"] [:div "D3"]]]
```

Here is an example where the (into) (concat) are just noise:

```clojure
(defn draw-graph [drawable mutable-graph force-layout]
  (let [{:keys [nodes paths]} @drawable]
    (into
     [:svg 
      (concat
       (for [path paths]
         [blink path mutable-graph force-layout nodes])
       (for [[node idx] (map vector nodes (range))
             :when (not (vector? (:id node)))]
         [bnode node idx mutable-graph force-layout])))))
```

Related discussion:
https://github.com/r0man/sablono/issues/40

Hiccup ignores the extra list generated, so this change would be both more convenient, and closer to the behavior of hiccup.

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -304,7 +304,6 @@
     (.apply (.' js/React :createElement) nil
             (reduce-kv
              (fn [a k v]
-               (prn "V" v "nested?" (nested? v))
                (when (>= k first-child)
                  (if (and (not (string? v))
                           (not (and (vector? v) (not (nested? v))))
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -28,7 +28,7 @@
 
 (defn valid-tag? [x]
   (or (hiccup-tag? x)
-      (ifn? x)
+      (fn? x)
       (instance? NativeWrapper x)))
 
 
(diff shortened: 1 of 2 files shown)


mainej:
:+1: for splicing in sequences

snoe:
:+1:

GetContented:
This may bring in subtle issues around collision of keys when you're needing to uniquely key items in a for or map. So long as one "namespaces" one's keys - so to speak - there shouldn't be any issue. (ie writes ^{:key (str "some-name" id)} rather than ^{:key id}).

GetContented:
Also, @timothypratley, (not really) speaking of turtles or state machines, `lambda` has a `b` in it.

holmsand:
While I agree that it would be nice to not have to supply keys, this patch would change the performance characteristics of Reagent fundamentally.

React uses the key attribute to avoid re-rendering of components when their order changes (see for example https://facebook.github.io/react/docs/multiple-components.html). React uses arrays to signal that a bunch of components are "dynamic", and so can be re-ordered – in Reagent we use `seq`s for the same thing.

We *could* generate keys automagically when they're not present – that would get rid of the warnings from React. But then it would be very easy to forget the keys, and very hard to tell if you have keys in the right place. I still think it is better to warn about missing keys always, even if it is slightly less convenient...

timothypratley:
Yeah this is a bad idea.

PR #198, opened 2015-10-28
Support for registering custom components

### LINKED ISSUE #48: Custom tags
When doing server side rendering, to use components you have to manually traverse the generated html and enhance the special nodes, as you cannot return a reference to a fn from the server.

Example:
(defn my-component [n]
  [:span (str n)])

Returned from the server:
[:div [:my-component 3]]

From there I manually scan it and replace :my-component for the fn reference.

Something like this would solve the problem and avoid to scan:
(r/register-component! my-component :my-component)

galdolber:
Fixes #48

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,4 @@
+{:paths ["src/reagent"]
+ :deps {org.clojure/clojure {:mvn/version "1.7.0"}
+        org.clojure/clojurescript {:mvn/version "1.7.48"}
+        cljsjs/react {:mvn/version "0.13.3-1"}}}
--- a/src/reagent/core.cljs
+++ b/src/reagent/core.cljs
@@ -294,3 +294,14 @@ the result can be compared with ="
   ;; with future versions of React (and return nil). 
   [c]
   (comp/component-path c))
+
+(defn register-component
+  "Creates an alias for a component.
+
+  (r/register-component :my-component my-component)
+
+  Can be used later as:
+
+  [:my-component]"
+  [name component]
+  (tmpl/register-component name component))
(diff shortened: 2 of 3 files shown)


galdolber:
This was taken from https://github.com/k2052/reagent/commit/9a6ab3ba9a833868148aeb9a98e1d1a77ed43da5

awb99:
Will this be merged eventually?

Deraen:
The feature will probably be added, but the implementation will be different than one in this PR.

mike-thompson-day8:
@Deraen  I'm getting close to needing this feature in re-frame in a future release.  I was wondering what your plans where in this area, and if there was anything I could do to help this along?

Deraen:
@mike-thompson-day8 My plan is to use Compiler interface to allow replacing the method used to create elements, so user can handle custom tags there. Could work even now by creating custom compiler with `make-element` which handles certain keywords, something like:

```
(defn make-element-custom [this argv component jsprops first-child]
  (let [component (case component :custom foobar-component component)]
    (tmpl/make-element this argv component jsprops first-child)))
```

Though `make-element` has to return React element directly, so that could be too low level for this. As-element / vec-to-elem could be better level to handle Reagent elements.

https://github.com/reagent-project/reagent/blob/master/src/reagent/impl/template.cljs#L296

But the Compiler interface and tmpl functions aren't (yet) part of the public API, so they could still change.

This isn't very simple, but I think this will be advanced feature and it doesn't need to be simple to use.

I'd also like to hear what is the usecase in Re-frame from custom elements.

Deraen:
Lets continue on a new issue: https://github.com/reagent-project/reagent/issues/503

Issue #191, opened 2015-10-07
API documentation GitHub page


yogthos:
since API docs are being generated with codox already it would be nice to add a page listing the API

jborden:
One small suggestion along these lines: It would be convenient if https://reagent-project.github.io/index.html had a direct link to the API docs in the top header. Looks like there would be room for it after the 'GITHUB' link.

yogthos:
👍

PR #203, opened 2015-11-04
Updated codox to the latest one

ducky427:
I've updated codox to the latest one. I've also generated the codox and put them [here](https://ducky427.github.io/reagent-docs). Hopefully the generated docs can live under reagent-project GH namespace.

Also see #191.


atroche:
Seems like this can be closed now, because the codox dep is already at 0.9.0

Deraen:
I'll add docs to http://reagent-project.github.io/ at some point.

PR #207, opened 2015-11-26
Fix doc formatting

sir-pinecone:
Just fixing up some indentation.

### DIFF
--- a/src/reagent/core.cljs
+++ b/src/reagent/core.cljs
@@ -18,17 +18,16 @@
 (defn create-element
   "Create a native React element, by calling React.createElement directly.
 
-That means the second argument must be a javascript object (or nil), and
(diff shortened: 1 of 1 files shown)


Deraen:
While some docstrings are clearly indented badly, in some cases you are just changing the indentation style.

It is matter of opinion how to indent the docstrings. Afaik most editors (Emacs, Vim) align multi-line strings (including docstrings) to the column where " is on the first line: https://github.com/bbatsov/clojure-style-guide#align-docstring-lines. Also, clojure.core uses this style.

sir-pinecone:
@Deraen Ah yes, that's a good point. I'll fix this up and revert the indentation changes that go against the clojure.core style.

sir-pinecone:
OK, fixed up my change.

holmsand:
Thanks!

PR #221, opened 2016-02-27
update readme version of react-with-addons

runningskull:


jwhitlark:
I've now hit this bug twice.  Can we please merge this in?

PR #233, opened 2016-04-16
Upgrade to latest React

tatut:
I tried upgrading React to the latest version. It should only contain internal changes.
All examples on the project page worked fine.

I ran tests on my machine with latest Chrome, Safari and Firefox, no errors:

Ran 93 tests containing 2034 assertions.
0 failures, 0 errors.


mike-thompson-day8:
@tatut thanks!  This will soon be a part of a new alpha release and available in Clojars for all to further test.

PR #236, opened 2016-04-27
Update Cljsjs React dependency

Deraen:
The latest Cljsjs React build adds missing nativeEvent property for
SyntheticEvent object on the extern file.

https://github.com/cljsjs/packages/commit/0db9cd856e6d5e35f275811225143233bc43048b


mike-thompson-day8:
Thanks for the vigalence @Deraen

PR #239, opened 2016-05-17
Add issue template

danielcompton:
This gets shown to every user as they create an issue in GitHub which might be more prominent than the CONTRIBUTING.md banner.


mike-thompson-day8:
Oh, this is excellent, thanks

PR #243, opened 2016-05-23
Create LICENSE

dijonkitchen:


mike-thompson-day8:
The structure of the copyright line should be:
`Copyright (c) <year> <copyright holders>`

This PR proposes a licence file which does not nominate the "copyright holders" part.  Before this PR could be accepted, Dan would have to nominate who he wants to hold the copyright (presumably himself).

dijonkitchen:
Opinions @holmsand ?

yarwelp:
What's the "blocked" label mean? Is it just that the original PR is lacking the name and year or is there some other issue also?

gadfly361:
@yarwelp This is blocked because we need guidance from the original author, @holmsand. 

This is his project and even though it is likely safe to assume the copyright holder would indeed be him, it is important to not copyright something on someone else's behalf.

dijonkitchen:
Given this is so long-lived, would it be easier to just add his name and merge this? Or shall we close it?

Deraen:
Something like this should be a safe bet:

```
Copyright (c) 2013-2017 Dan Holmsand
Copyright (c) 2017 Reagent maintainers
```

Would be indeed good to have @holmsand approval, but he already holds copyright for his work (IANAL, but it happens automatically), and the notice is only for documentation purposes. So I think it would be safe to add this?

Deraen:
Btw. jQuery has interesting notice on top of MIT license: https://github.com/jquery/jquery/blob/master/LICENSE.txt

```
Copyright JS Foundation and other contributors, https://js.foundation/

This software consists of voluntary contributions made by many
individuals. For exact contribution history, see the revision history
available at https://github.com/jquery/jquery

The following license applies to all parts of this software except as
documented below:
```

But line mentioning `Reagent contributors` should have the same effect.

dijonkitchen:
See #318. Need to close because of old branch.

PR #250, opened 2016-07-09
Fixes require typo in readme example

danprince:
Just a tiny conflict between importing as `r` but using as `reagent` in the first example.


gadfly361:
Thanks!

PR #260, opened 2016-09-13
Fix #259: Call original ref in ReagentInput

### LINKED ISSUE #259: ReagentInput `:ref` broken
https://github.com/reagent-project/reagent/commit/42ba1662cbc62dc7ebc0c437abe990f1acbe9840 broke using `:ref` attribute with input elements. This is because `:ref` attribute is always overriden for these elements: https://github.com/reagent-project/reagent/blob/42ba1662cbc62dc7ebc0c437abe990f1acbe9840/src/reagent/impl/template.cljs#L186

If the code keeps on going using `:ref` callback, it must keep track of the original function and call that. If ref callback is used, string ref can never be supported here, but maybe that is not a problem as those haven't really worked with Reagent anyway.

Deraen:


holmsand:
Hmm, thinking a bit more about this patch – I think there is a slight problem here. This will give React a new function for the ref on every render, which means that React will call it twice on rendering, as described in https://facebook.github.io/react/docs/more-about-refs.html – even if the :ref passed to us doesn't change.

I guess the easiest way to fix this would be to go back to using findDOMNode for out purposes, and pass the ref on as is, but I'll have to think about that a bit more.

Deraen:
Yes, I noticed this too, but choose to ignore that as I couldn't find solution to this while using `:ref`.

Using `dom-node` on `input-set-value` should be quite easy way to fix this.

holmsand:
Yes, it is probably the right way. The only downside is that we'll have to add back some code to make sure that we don't try to access the input after the component is unmounted, but that's quite easy to do.

The alternative would be to save the ref passed to us, and the function we create, and check if the ref changed on the next render, but that seems messier to me right now.

src/reagent/impl/template.cljs
```diff
@@ -175,15 +175,14 @@
     (let [v ($ jsprops :value)
           value (if (nil? v) "" v)
           on-change ($ jsprops :onChange)]
-      (when (nil? ($ this :cljsInputElement))
+      (when (nil? ($ this :cljsDOMValue))
```
Deraen: Not completely sure if this is good way to check if this is the first render.

Deraen:
Updated with my take on this. Not sure on the one `if` condition on if the input is being rendered the first time.

Is there a case where this will access the component after it is unmounted?

src/reagent/impl/template.cljs
holmsand: Should be OK, since value could not be nil.

holmsand:
Yes, input-set-value will be called after the next rendering when on-change is called, and the rendering may cause the component to become unmounted. So, a component-will-unmount will be needed (setting cljsDOMValue to nil).

Another thing: I think I would prefer to just require "react-dom" in one place (i.e. reagent.dom), so that dependence on react-dom can be avoided when using React native. That will require some trickery to avoid circular dependencies, though...

I'll take a better look later.

Deraen:
I think the best way to solve react-dom dependency is to load it in util ns.

holmsand:
@Deraen Loading react-dom in util doesn't really work, since cljsjs.react.dom is loaded by reagent.dom. And moving the requiring of cljsjs.react.dom to util would force it to be loaded even when it isn't needed (i.e React native).

Instead I've made reagent.dom provide a `find-dom-node` function to template.cljs when reagent.dom is loaded, which should work since we can't very well have dom inputs without the dom...

I've also added a check that the input isn't unmounted before accessing the input element. Take a look.

Deraen:
Hmm, I didn't account for foreign library load order & react-native case.

Anyways, working for me now!

Deraen:
@holmsand Could you release a version with the fix? Or even a snapshot would be great.

PR #263, opened 2016-09-20
Feature proposal: add reaction that allows the user to explicitly notify watchers

grav:
This pull request is an extension to the api `reagent.ratom/make-reaction-notify`, that allows creation of reactions that allow for explicit notification of watchers. 

This way, it is possible to create a reaction that for instance wraps asynchronicity:

```
(def input (r/atom 42))

(def squareroot-of-input (reagent.ratom/make-reaction-notify (fn [notify] (-> (js/fetch (str "http://heavymath.com/squareroot/" @input))
                                                      (.then notify)))))

(def result (reaction (* 2 @squareroot-of-input)))

(defn view []
   [:p (str "The double square-root of " @input " is " @result)])

```

Here, the `squareroot-of-input` reaction will fire when `input` changes, and then notify the `result` reaction when the xhr request is done.

It's possible to call `notify` several times in a function. It's not a new feature for FRP libraries (I've used it in RxJS and Reactive Cocoa), I just couldn't find anything similar in Reagent.

The implementation is very crude, so it's more a kind of proposal for a feature than a final implementation.

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -493,6 +493,25 @@
     reaction))
 
 
+(defn make-reaction-notify [f & args]
+  (let [reaction (apply make-reaction f args)]
+    ;; override _run method of Reaction type so that it calls f with the `notify` function
+    (set! (.-_run reaction) (fn [check]
+                                 (let [oldstate (.-state reaction)
+                                       notify (fn [res] (when-not (.-nocache? reaction)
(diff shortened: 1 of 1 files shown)


Deraen:
Closing this 8 year old proposal. Now, looking at this idea, it does seem interesting in some cases. Some other features in React (`useEffect` etc) also can be used do deal with similar cases.

I've opened a issue to document this, in case I want to come back to consider this later: #616

PR #266, opened 2016-10-06
Fix react-with-addons example

Deraen:
Reagent now depends on `cljsjs/react-dom` and `cljsjs/react-dom-server` so
direct dependency to `cljsjs/react` won't work.

Btw. if wanted, the old way could be enabled by adding direct dependency to `cljsjs/react` to Reagent, and adding `cljsjs/react` exclusions to dom and dom-server deps on Reagent side. My opinion is to not do tricks here and leave this to the user. Using global exclusions is probably the best way to do this, because other libs beside Reagent might anyway have transitive dependency to `cljsjs/react` and global exclusions  takes care of those also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,16 @@ To use Reagent in an existing project you add this to your dependencies in `proj
 
 This is all you need to do if you want the standard version of React. If you want the version of React with addons, you'd use something like this instead:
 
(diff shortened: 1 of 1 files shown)


Deraen:
I was wrong. The existing example works.

PR #270, opened 2016-10-20
Add test showing Reagent running disposed reactions

danielcompton:
While fixing https://github.com/Day8/re-frame/pull/259 we noticed that Reagent was happily running reactions that had been disposed of. This commit adds a failing test exposing that behaviour. I'm not sure exactly what the right behaviour is here, but I don't think Reagent should silently allow this.


Deraen:
I'm finally taking a look at this, as I'm trying to merge or close all extra PRs.

It would seem reasonable that Reactions can't be "forced to run" after disposing of them. But I also wonder if this is such a low-level use that we want to leave this to the users to avoid.

But OK, `dispose!` is exposed in core ns.

At the very least, we should document the expected behaviour here.

Deraen:
The test case is merged, with broken assertion commented out. Further decisions on #617.

Issue #265, opened 2016-10-05
Supporting controlled custom inputs components


madvas:
Hi,

Speaking for all users of [cljs-react-material-ui](https://github.com/madvas/cljs-react-material-ui) with reagent. We hit the problem, that reagent's caret repositioning hack doesn't work for [TextField component](http://www.material-ui.com/#/components/text-field).

I prepared very simple demonstration [project](https://github.com/madvas/reagent-material-ui), which is basically fresh reagent template with cljs-react-material-ui. After bit of a hacking with reagent code, we figured out if we overwrite `reagent.impl.template/input-component?` like [this](https://github.com/madvas/reagent-material-ui/blob/4aba539b27bec2cde5d87ee3abc52dc43cea2fe0/src/cljs/reagent_material_ui/core.cljs#L16-L21) we can make repositioning work. That last condition is how to reliably identify TextField component, even after advanced optimisations. 

However this hack has undesired side-effect. We can no longer set value programmatically, only by typing with keyboard. As you can see [here](https://github.com/madvas/reagent-material-ui/blob/4aba539b27bec2cde5d87ee3abc52dc43cea2fe0/src/cljs/reagent_material_ui/core.cljs#L25-L25) if you uncomment line, you'll see value doesn't change for `ui/text-field`, but for `:input` it works ok. 

This way I'd like to kindly ask you for an advise how could we make repositioning for TextField work. I don't see too much into repositioning internals, not sure if we can somehow hack TextField component, or changes to Reagent will be needed.

Thank you very much

Deraen:
#282 is merged and released as 0.8.0-alpha2

metametadata:
Although I'm a bit late to the party as the fix is already merged, I have 2 concerns with it:

1) The new API isn't documented anywhere at the moment. The example of using the new hooks looks quite [scary](https://github.com/madvas/cljs-react-material-ui/issues/17#issuecomment-268807607).
2) I suspect there's a possibility to provide the universal solution without touching Reagent core and making the API much simpler, smt. like:

```clj
(def text-field (reagent-input-fixer/fix 
                              (r/adapt-react-class (aget js/MaterialUI "TextField"))))
```

This can be implemented as a new lib or as an utility function in Reagent.

[...]

PR #282, opened 2016-12-23
Support synthetic text input components

arbscht:
Submitting PR for review and discussion at this stage.

**Problem:** Reagent controls text input components to handle updates and caret repositioning correctly, but this assumes that only `<input>` elements are rendered by text input components in Hiccup. This excludes Reagent users that, for example, wrap their `<input>` element in a `<div>` or otherwise have arbitrary behaviour/structure. The result is that these components have caret positioning bugs, as seen in #265.

**Proposed solution:** Add more affordances to Reagent's input handling API so that "synthetic" input components (i.e. not just rendering `<input>` at the root) can inform Reagent of how to locate the correct input element to manage within its DOM tree; and what to do when values change.

In particular, in `template.cljs`:

* Take a map of options with `r/adapt-react-class` with optional overrides for input handling where applicable.
* Allow `input-component?` to detect synthetic inputs as input components needing caret control on user interactive updates.
* Allow `input-set-value` to be intercepted by synthetic inputs so they can refine which `node` is the correct input element inside their DOM subtree.
* Add a hook for `on-write` so synthetic inputs can respond appropriately to manage their special local state when new values are meant to be written to the component.
* Allow intercepting `on-change` so synthetic inputs can transform values or sync state.

These were all motivated by #265 so that cljs-react-material-ui could get its TextField component working with Reagent's input handling, but the changes are meant to be agnostic of client quirks. You can see an example of usage in this snippet: https://github.com/madvas/cljs-react-material-ui/issues/17#issuecomment-268807607

How does this approach look?

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -112,55 +112,68 @@
   [input-type]
   (contains? these-inputs-have-selection-api input-type))
 
-(defn input-set-value [this]
+(declare input-component-set-value)
+
+(defn input-node-set-value
+  [node rendered-value dom-value component {:keys [on-write]}]
+  (if-not (and (identical? node ($ js/document :activeElement))
+            (has-selection-api? ($ node :type))
+            (string? rendered-value)
+            (string? dom-value))
+    ;; just set the value, no need to worry about a cursor
+    (do
+      ($! component :cljsDOMValue rendered-value)
+      ($! node :value rendered-value)
+      (when (fn? on-write)
+        (on-write rendered-value)))
+    
+    ;; Setting "value" (below) moves the cursor position to the
+    ;; end which gives the user a jarring experience.
+    ;;
+    ;; But repositioning the cursor within the text, turns out to
+    ;; be quite a challenge because changes in the text can be
+    ;; triggered by various events like:
+    ;; - a validation function rejecting a user inputted char
+    ;; - the user enters a lower case char, but is transformed to
+    ;;   upper.
+    ;; - the user selects multiple chars and deletes text
+    ;; - the user pastes in multiple chars, and some of them are
+    ;;   rejected by a validator.
+    ;; - the user selects multiple chars and then types in a
+    ;;   single new char to repalce them all.
+    ;; Coming up with a sane cursor repositioning strategy hasn't
+    ;; been easy ALTHOUGH in the end, it kinda fell out nicely,
+    ;; and it appears to sanely handle all the cases we could
+    ;; think of.
+    ;; So this is just a warning. The code below is simple
+    ;; enough, but if you are tempted to change it, be aware of
+    ;; all the scenarios you have handle.
+    (let [node-value ($ node :value)]
+      (if (not= node-value dom-value)
+        ;; IE has not notified us of the change yet, so check again later
+        (batch/do-after-render #(input-component-set-value component))
+        (let [existing-offset-from-end (- (count node-value)
+                                         ($ node :selectionStart))
+              new-cursor-offset        (- (count rendered-value)
+                                         existing-offset-from-end)]
+          ($! component :cljsDOMValue rendered-value)
+          ($! node :value rendered-value)
+          (when (fn? on-write)
+            (on-write rendered-value))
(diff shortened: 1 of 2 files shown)


edannenberg:
Is there anything we can do to help getting this merged?

alvatar:
This is a pretty big bug affecting https://github.com/madvas/cljs-react-material-ui/issues/17#issuecomment-268807607, making the text fields almost unusable. I'll be happy to help with this too.

superstructor:
This is a serious bug. Text fields are pretty fundamental to any application. What are the next steps to get this merged as soon as possible please ? @Deraen

alvatar:
Yes please.

alvatar:
Wait, this PR is from 2016. Why is this issue neglected? It doesn't make sense. This is a pretty serious bug.

arbscht:
> Why is this issue neglected? It doesn't make sense. This is a pretty serious bug.
I don't know if there's anything in the roadmap or architecture that is blocking this change. But it is relatively hairy and also will require some co-ordination with cljs-react-material-ui (or whatever client lib) to ultimately resolve for users. That said, I'm available and happy to work through reviews/make changes, answer questions, add docs/tests, etc. I rebased the PR recently so it should still be reviewable. Let us know @Deraen.

Deraen:
I do have limited time to work on Reagent, and I have chosen to work first on things that are high priority for myself (Foreign libs / Node modules support, new test runner, React 16). Currently things are pretty much blocked while I ponder when to release 0.8.0 with just the React dependency change, but after that I plan on merging new features.

superstructor:
@Deraen I understand you may have limited time, but we do not. @arbscht and myself would be paid to do the work necessary to get this merged. Others have also offered to help if nessecery. 

This is not a feature, this is a critical bug impacting large production users of reagent. It makes text fields unusable when reagent is used with some popular component libraries like Material-UI. 

It's also understandable that you have different immediate priorities for reagent, but for us this is the top priority. 

We are not asking for you to do the work. All we are asking as production users of reagent is that you trust some of the community enough to delegate getting this merged so that the whole community of users can benefit from bug-free text fields, and everyone doesn't have to maintain their own reagent fork. 

If it needs tests, more documentation, rebasing on other work, or anything else we will get whatever is needed done on our own time. We happy to minimise the impact on your time as much as possible.

Deraen:
Regarding release cycle, my original plan was to only release the dependency change as part of 0.8, but fixing some potential problems with that is taking longer than I hoped. If I were to now release beta or rc with current changes, I would probably wait ~2 weeks before proper release.

Alternative would be to include this and some other changes in 0.8. This would make for larger release, which I don't like much, but this fix would possibly be released faster (I could at least cut another alpha soon). Opinions?

src/reagent/impl/template.cljs
```diff
@@ -270,13 +342,20 @@
 (declare as-element)
 
 (defn native-element [parsed argv first]
-  (let [comp ($ parsed :name)]
+  (let [comp ($ parsed :name)
+        synthetic-input ($ parsed :syntheticInput)
+        synthetic-on-update ($ parsed :syntheticOnUpdate)
```
Deraen: Move `synthetic-on-update` and `synthetic-on-change` under the then branch of `if` where checking for `synthetic-input`, so that these aren't unnecessarily evaluated for each element.

arbscht: @Deraen Thanks for looking it over!

> But my point is, if I understand correctly, this change only affects `adapt-react-class`? Other cases should work already.

Yes I believe you're right, I could have been more specific, this is about `adapt-react-class`.

arbscht:
> Alternative would be to include this and some other changes in 0.8. This would make for larger release, which I don't like much, but this fix would possibly be released faster (I could at least cut another alpha soon). Opinions?

+1 for this option. A bigger release is not ideal, but a lot of us have been awaiting this bugfix for some time already. Also it would be suboptimal if this bugfix got entangled with even more high-cost upgrades in a future release, so earlier is better IMO.

alvatar:
oh yeah! Thanks @Deraen and everyone!

PR #301, opened 2017-06-01
Improve assert messages all around

jeaye:
This aims to steer away from error messages like this:

image


jeaye:
What's the schedule looking like for merging?

PR #306, opened 2017-07-12
Unified React require

Deraen:
ClojureScript 1.9.854 allows using the same require form for npm-deps, cljsjs packages and Node target.

```
;; cljs
(ns ... (:require [react-dom :as react-dom]))

(react-dom/render ...)

;; foreign-lib
:global-exports {react-dom ReactDOM}
;; Cljs call is compiled to
var a = window.ReactDOM;
a.render(...)
;; or something similar

;; :target :nodejs
var a = require("react-dom");
a.render(...)

;; npm-deps
module$foo$bar$react$react-dom.render(...)
;; where module$foo$bar$react$react-dom is name for converted Closure module
```

Version built using `:npm-deps` is online at https://reagent-project.github.io/reagent-site-npm-deps-test/. In this version the React code is optimized using Closure, which results in a bit smaller output: 463K -> 440K JS, non-gzipped.


Deraen:
This should be mostly ready now that Cljs 1.9.854 is out.

Unless someone has comments about this, I plan on merging this to master soon and releasing a SNAPSHOT.

I think next version (0.8) will include only this change, as this change will required the new Cljs version, and it will take some time for everyone to upgrade to that.

project.clj
```diff
@@ -93,7 +95,12 @@
                                        "examples/geometry/src"]
                         :compiler {:parallel-build true
                                    :main "reagentdemo.core"
-                                   :output-to "outsite/public/js/main.js"}}}}
+                                   :output-to "outsite/public/js/main.js"
+                                   :language-in :ecmascript6
+                                   :language-out :ecmascript3
+                                   :npm-deps {:react "15.6.1"
+                                              :react-dom "15.6.1"
+                                              :create-react-class "15.5.3"}}}}}
```
wodin: Why `:create-react-class "15.5.3"` when above you have `:dependencies [cljsjs/create-react-class "15.6.0-1"]`?

Deraen: Just a mistake.

Issue #272, opened 2016-11-06
Handle thrown exceptions


danielcompton:
React doesn't handle exceptions particularly gracefully out of the box. Exceptions thrown during rendering will bubble up and destroy all of the components up to the root, without calling lifecycle methods. https://github.com/facebook/react/issues/2461 is an issue that tracks this.

This exception handling behaviour affects us in re-frame. If app code throws an exception while rendering, then the component is destroyed without calling component-will-unmount, and Reagent doesn't decrement the watch count on the reaction (re-frame subscription). Because the watch count is inaccurate, Reagent never calls the on-dispose, and the subscription stays in the re-frame cache, causing further issues.

It looks like if we provide [`unstable_handleError`](https://github.com/facebook/react/blob/ef99e7e096cb661de0e72f93a442fbfeaaa07b1c/src/renderers/shared/stack/reconciler/ReactCompositeComponent.js) as a method on a Reagent component, then the initial mount is called in a try/catch. unstable_handleError is unstable, but this may be worth investigating, either now or later on when React marks it as stable.

pesterhazy:
Here's my attempt at creating an "error boundary" using the `unstable_handleError` API: https://gist.github.com/pesterhazy/d163a8b3f1f1c6a0dac235858776c14b

If used to wrap a root component as explained in the gist, the component gracefully handles exceptions thrown downstream in the component hierarchy.

Let me know if that works for you.

Deraen:
Would be great to have this improved in the next release.

pesterhazy:
This came up on clojurians the other day. Timothy mentioned that when he tested this recently, Reagent (and React) was able to recover from exceptions thrown in the render functions. I also wasn't able to reproduce the original problem anymore (but didn't try too hard), even without the error boundary set.

I tried to go through React's changlog to see if any recent changes could have fixed this but didn't find anything.

Is it possible that somehow recently this problem has been solved, either on React's or Reagent's side?

If the problem persists ("unable to read property of" or similar errors thrown after fixing an exception in a render method and rerendering), a minimal repro case would be very helpful.

danielcompton:
https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html is the way this will be handled in React 16. It seems like it should be pretty easy to support in Reagent.

Deraen:
[...]

PR #308, opened 2017-08-03
Update default React to version 16

Deraen:
Needs some fixes:

```
21 failures, 1 errors.
```

We should also use Error Boundaries:  #272  https://facebook.github.io/react/blog/2017/07/26/error-handling-in-react-16.html


dustingetz:
Is anyone using this branch successfully?

jeaye:
Looks like the official React 16 release was today, which is going to spark some more interest. An update on this would be great.

dustingetz:
haven't tried it yet but shouldn't it just work

Deraen:
Updated to final version

```
23 failures, 1 errors.
```

package.json
```diff
@@ -1,10 +1,10 @@
 {
     "dependencies": {
         "browser-resolve": "^1.11.2",
-        "create-react-class": "^15.5.3",
+        "create-react-class": "^15.6.0",
         "module-deps": "^4.1.1",
-        "react": "^15.5.3",
-        "react-dom": "^15.5.3",
+        "react": "^16.0.0",
+        "react-dom": "^16.0.0",
```
YurySolovyov: why not strict?

Deraen: As these have always been like this, and I don't know much about Node best practices.

So it would be best to not usre any specifiers, ~ or ^ etc.?

YurySolovyov: Well, on the `:npm-deps` you have them strict, so at least for consistency.
I think it would be better to have a bit more determinism by omitting the `^`

seako: best to not use any specifiers imo. that way one can test against a known good configuration and know that users will get that same configuration.

Deraen:
No hurry on this one - Reagent already supports React 16 - this one just changes the default React version and removes separate test cases for React 16.

grav:
@Deraen Do you have a working example of React 16 with Reagent?

I've tried using Reagent 0.8.0-alpha2 with React 16.0.0-0, and a minimal example:

```
(defn app []
  [:div "hello"])

(reagent.core/render [app] (.getElementById js/document "app"))
```

 throws an error 
```
react-dom.inc.js:15986 Uncaught TypeError: getNextDebugID is not a function
    at ReactCompositeComponentWrapper.instantiateReactComponent [as _instantiateReactComponent] (react-dom.inc.js:15986
...
```

I've verified that I am actually running React 16 from the js console:
```
> React.version
< "16.0.0"
```

Deraen:
The demo app should work:

```
cd test-environments/browser-umd-react-16
lein figwheel
```

Check that you have correct version of all cljsjs React packages using `lein deps :tree` / `boot show -d`.

grav:
Yes, sorry, I only had `cljsjs/react`, not `cljsjs/react-dom` for React 16. Now it works!

YurySolovyov:
FWIW [React 16.1.0 is released](https://github.com/facebook/react/releases/tag/v16.1.0)

PR #314, opened 2017-10-06
Add IWithMeta protocol to RAtom

### LINKED ISSUE #257: Ratom missing 'with-meta'
Because `meta` [exists for ratoms](https://github.com/reagent-project/reagent/blob/master/src/reagent/ratom.cljs#L149), I would also expect `with-meta` to be there.

Here's a reference to the clojurescript core source : [with-meta example](https://github.com/clojure/clojurescript/blob/709cc57d003590945329ea8745b799e75ac4ba88/src/main/cljs/cljs/core.cljs#L5838)

frerom:
Fixes #257

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -146,6 +146,9 @@
   (-swap! [a f x y]      (-reset! a (f state x y)))
   (-swap! [a f x y more] (-reset! a (apply f state x y more)))
 
+  IWithMeta
+  (-with-meta [_ new-meta] (RAtom. state new-meta validator watches))
+
   IMeta
   (-meta [_] meta)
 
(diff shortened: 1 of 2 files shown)


danielcompton:
This would return a new instance of the Ratom when calling `with-meta`. This seems idiomatic amongst other core Clojure code, but I'm not sure what its effects on Reagent would be, both because it makes heavy use of meta, and because object identity is quite important in Reagent, and could lead to confusing scenarios. It might be fine, but this should probably be checked out further.

Deraen:
I think this will be safe. Reagent doesn't change RAtom meta ever, so only case where new RAtom is created using `with-meta` should be when user does so. And I think it won't be easy to do that accidentally, unless one has two nested RAtoms or such.

```clj
(deftest ratom-with-meta
  (let [value {:val 1}
        meta-value {:meta-val 1}
        state (r/atom (r/atom value))]
    (is (= (meta @state) nil))
    (reset! state #(with-meta % meta-value))
    (is (= (meta @state) nil))   
    )) 
```

PR #325, opened 2017-11-18
Fix aget/aset use with objects

### LINKED ISSUE #324: Fix aget used with objects
`aget` should only be used with arrays.

Deraen:
This removes the interop macros which used aget/aset and prevented
Closure naming mangling, now normal property and method access is used
where possible, and goog.object used when using variable keys. Further
testing is needed to ensure this works correctly with Closure, as there
are some properties that are used with goog.object in one place, and as
property in another.

Fixes #324 

This still throws several assertion errors, caused by `aset` currently checking that index is inside existing array, though JS arrays automatically grow when assigning values to new indices: `Assert failed: (< idx (alength array))`


mfikes:
FWIW, I asked about assigning new indices via `aset` and David Nolen's view is that something like
```
(let [a (into-array [0 1])] 
  (aset a 2 2) 
  a)
```
is not a correct program, but he he admits that it is also sufficiently host-y where it's a bit of an edgecase.

Issue #339, opened 2018-02-04
Tests fail without global karma install


alex-dixon:
#### Version
0.8.0-alpha2

#### Description
Tests fail on systems with no global `karma` command

```bash
./run-tests.sh    
##
## TESTING browser-node
##

+ lein do clean, doo chrome-headless test once

;; ======================================================================
;; Testing with Chrome-headless:

[doo] Failed to run command:  #error {
 :cause "error=2, No such file or directory"
 :via
 [{:type java.io.IOException
   :message "Cannot run program \"karma\": error=2, No such file or directory"
   :at [java.lang.ProcessBuilder start "ProcessBuilder.java" 1048]}
  {:type java.io.IOException
   :message "error=2, No such file or directory"
   :at [java.lang.UNIXProcess forkAndExec "UNIXProcess.java" -2]}]
 :trace
 [[java.lang.UNIXProcess forkAndExec "UNIXProcess.java" -2]
  [java.lang.UNIXProcess <init> "UNIXProcess.java" 247]
  [java.lang.ProcessImpl start "ProcessImpl.java" 134]
  [java.lang.ProcessBuilder start "ProcessBuilder.java" 1029]
  [java.lang.Runtime exec "Runtime.java" 620]
  [java.lang.Runtime exec "Runtime.java" 485]
  [doo.shell$exec_BANG_ invokeStatic "shell.clj" 50]
  [doo.shell$exec_BANG_ invoke "shell.clj" 43]
  [doo.shell$sh invokeStatic "shell.clj" 76]
  [doo.shell$sh invoke "shell.clj" 71]
  [doo.core$run_script invokeStatic "core.clj" 268]
  [doo.core$run_script invoke "core.clj" 236]
  [user$eval7316$fn__7332 invoke "form-init4215627720157311738.clj" 1]
  [clojure.core$map$fn__4785 invoke "core.clj" 2644]
  [clojure.lang.LazySeq sval "LazySeq.java" 40]
  [clojure.lang.LazySeq seq "LazySeq.java" 49]
  [clojure.lang.RT seq "RT.java" 521]
  [clojure.core$seq__4357 invokeStatic "core.clj" 137]
  [clojure.core$every_QMARK_ invokeStatic "core.clj" 2570]
  [clojure.core$every_QMARK_ invoke "core.clj" 2563]
  [user$eval7316 invokeStatic "form-init4215627720157311738.clj" 1]
[...]

PR #340, opened 2018-02-04
Remove global karma dependencies

alex-dixon:
#339

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,7 +6,6 @@ machine:
 dependencies:
   override:
     - lein deps
-    - npm install -g karma-cli
     - ./prepare-tests.sh
   cache_directories:
     - test-environments/browser-node/node_modules
--- a/project.clj
+++ b/project.clj
@@ -40,6 +40,8 @@
              :css-dirs ["site/public/css"]
              :repl false}
 
+  :doo {:paths {:karma "./node_modules/.bin/karma"}}
+
   ;; No profiles and merging - just manual configuration for each build type
   :cljsbuild
   {:builds


danielcompton:
`karma-cli` is a wrapper around the local Karma that is installed, you can have it globally installed and still use many different versions of the `karma` library in different projects. Would it be sufficient to add documentation to explain what is required here?

alex-dixon:
Of course! I thought this might be nicer but that's just me. 

I didn't realize `karma-cli` was a wrapper. Surprisingly that doesn't seem to affect my thinking much.
It's still one less step/thing to read for contributors. `karma-cli` as dependency. Even though it's not an explicit one it's still there I think -- you could end up with different versions of it on devs machines and across CI runs, which could produce inconsistent test results for no apparent reason.

danielcompton:
> Even though it's not an explicit dependency it's still there I think -- you could end up with different versions of it on devs machines and across CI runs, which could produce inconsistent test results for no apparent reason.

karma-cli is meant to be a very stable wrapper that doesn't change much, it's only 80 lines of JS, with the last [release](https://github.com/karma-runner/karma-cli/releases) in 2016. There should be basically no scenario where you get inconsistent test results as it is just a launcher. See https://github.com/bensu/doo/pull/46 for more details on why using `karma-cli` is preferable, particularly around Windows support.

I think it would definitely be good for the development docs to mention installing `karma-cli`, I didn't see it mentioned on a quick scan through.

alex-dixon:
I can see you’re a seasoned expert on these matters. 

A test runner that’s downloaded 4 million times a week and requires the installation of a second package for command line support was unexpected for me. I’m sure there’s a lesson and a story or two in there somewhere. 

Nondeterminism in resolving npm packages bothers me very much. If I read the code correctly it says “resolve karma relative to the pwd (wherever that is) and if there’s a karma there then great! Let’s use that, whatever version it happens to be. ” If it doesn’t find a karma in this non deterministic way, it doubles down and looks for a globally installed version. If that’s found, great! Use it, whatever version it happens to be. 

The only thing I can think would aid windows support is a call to [path.normalize](https://nodejs.org/api/path.html#path_path_normalize_path) [here](https://github.com/karma-runner/karma-cli/blob/3be8633c616faae78fbb4cc841b9b19636c5a075/bin/karma#L76) which enforces platform specific slashes and the regex [here](https://github.com/karma-runner/karma-cli/blob/3be8633c616faae78fbb4cc841b9b19636c5a075/bin/karma#L54) which appears to do the same thing. 

Maybe ‘doo’ wasn’t/isn’t performing that step?

Deraen:
We already use locked down version of Karma (and plugins), see [package.json](https://github.com/reagent-project/reagent/blob/master/package.json). We should just document that `karma-cli` should be globally installed befory trying to run the tests.

Deraen:
Added mention about this to dev docs: https://github.com/reagent-project/reagent/commit/ae532864ce64ec4a7d4030611b44507ff43cd893

Deraen:
@alex-dixon Do you agree the documentation is enough? This and the issue can be close?

alex-dixon:
Sounds good. Thank you for the informative discussion @danielcompton @Deraen!

Issue #302, opened 2017-06-04
Add a guide to the docs


pesterhazy:
Reagent has goods docs but they're somewhat hard to find. Some topics are covered in the API docs, others are discussed in the NEWS posts. Some information again is best covered in re-frame's docs or in the Reagent Cookbook.

To make Reagent more accessible, it would be helpful to include a Table of Contents in the README that links to the various docs. Here's a breakdown of topics that I think would be useful to cover.

Thoughts?

## Mounting components

- r/render
- considerations to ensure proper reloading

## State management

- ratom
- reaction
- cursor
- track
- track!
- wrap

## React interop

- create-element
- adapt-react-class
- `[:> ...]`
- as-element
- reactify-component
- requiring react - how React is found

## Performance

- async rendering
- flush
- force-update-all
- force-update
- next-tick
- after-update

## Lifecycle methods

- List of available lifecycle methods
- Form-2 components vs Form-3 components
- Accessing props in Form-2 components
- Cleaning up: `with-let`
- How should-component-update works in Reagent

## Accessing DOM nodes/backing instances

- dom-node
- callback refs

## Render methods

- Accessing props and children
- Form-2 components: remembering to repeat arguments in the inner function
- `:render` vs `:reagent-render`

## Reagent-flavored hiccup

- Nested element syntax: `[:li.foo>a.bar]`
- Handling nil
- Vectors and sequences are handled differently
- Settinging keys through meta-data
- How props case is handled (`:on-change` -> `"onChange"`)

## Handling input elements

- Keeping cursor position
- Latency hacks
- Using 3rd party input elements (e.g. from Semantic UI React)

## Component-local state

- Component-local ratoms
- Form-2 components
- reagent.core/state and reagent.core/set-state

## Advanced topics

- Server-side rendering

gadfly361:
@pesterhazy I really like this idea!  Would you be interested in starting a PR for this?

There are several outstanding issues that I think this could help solve: [249](https://github.com/reagent-project/reagent/issues/249), [209](https://github.com/reagent-project/reagent/issues/209), [166](https://github.com/reagent-project/reagent/issues/166), and [147](https://github.com/reagent-project/reagent/issues/147)

I am hopeful that we can triage the relevant links to add to it. And if certain bullets are without links, it could potentially be a good opportunity for community members to write a blog post and help with some documentation.

xaris:
Hey @pesterhazy @gadfly361 ,

[...]

Issue #249, opened 2016-06-27
Documentation doesn't define fundamental concepts


vspinu:
One "feature" of the existing documentation is that it doesn't pay much attention to defining basic concepts. 

A couple of examples. 
1. What is `component` in reagent context? Is it a function or the returned hiccup data structure or something else?  
2. What is `mounted component` and how it differs from `component`?
3. In terms of the syntax, what is the difference between  `(fn [] [:div (some-component)])`, `(fn [] [:div [some-component])`, `(fn [] [:div some-component])`, or just `[:div [some-component)]`?  There is only a quick note concerning the first two: "The only difference would have been performance, since ”real” Reagent components are only re-rendered when their data have changed". 
4. What is `"real" component`?
5. How do you define `component's data`? 
6. How components are re-generated is only tangentially explained. There are two references on this but both are, at best, incomplete definitions: 
   - "Any component that uses an atom is automagically re-rendered when its value changes".  
     1. Does this mean that if I use one atom per application then all components will be re-rendered even on a small change of the atom's hashmap? So, I should use multiple attoms whenever possible?
     2. Does `uses an atom` mean that "the function that generates hiccup data structure derefs an attom"? I infer from the examples that the answer is "yes". That is, whenever value of an atom changes, all functions that deref that atom within their body will be called again to re-generate the hiccup data structures.
   - "The change could come from a deref’ed atom, the arguments passed to the component or component state". 
     1. I guess `arguments` are the values passed to functions which produce hiccup structures.  `deref'ed atom` is as in the above `uses an atom` idiom. But what is `component state`? 

It could be that the docs assume knowledge of `react` and maybe all the above have unique implicit definition.  Sorry if that's the case; my knowledge of react is close to zero.

I would expect, for a knowledgeable person, it would cost a tiny amount of time to create a small wiki glossary with exact definitions of the above concepts. I think It would dramatically improve the accessibility of the package for newbies like me. Thank you very much.

vspinu:
Have just discovered that [re-frame](https://github.com/Day8/re-frame)'s main page and [wikis](https://github.com/Day8/re-frame/wiki) clarify most of the above questions in a most fluent and informative way possible.

mars0i:
[reagent-cookbook](https://github.com/reagent-project/reagent-cookbook) also does a pretty good job of introducing basic concepts.

gadfly361:
I think the recent docs that were moved from re-frame and #343 are sufficient to close this.

Issue #209, opened 2015-12-07
Failures In Documentation


mike-thompson-day8:
There's been a recent thread on reddit containing stern criticism of Reagent (which, oddly, started as a rant against OM.Next). 
https://www.reddit.com/r/Clojure/comments/3vk58p/a_rant_on_om_next/

The most vociferous criticism was from "unknown4242" who was **certain** that reagent encouraged a ratoms-everywhere, "mutable" architecture - at least that's the impression I distilled from his slightly all over the place rant across multiple comments.  

So "unknown4242" has hurt himself using Reagent.  

At any scale above the trivial, that ratoms-all-over-the-place architecture will be problematic, and lead to a messy disaster.  That's what he experienced.

To me, it seems so obvious that this is a bad idea that it wouldn't even occur to me that anyone would do it.  But, after reading his rants, I looked back over the reagent docs and noticed that there is no advice anywhere to stop a person from doing this.  In fact, just the opposite, the examples, in attempting to keep things simple, use the multiple r/atoms approach. 

So I guess I can understand how an inexperienced person could get the wrong impression. Go in the wrong direction. Hurt themselves.

So this ticket is about us improving the docs.   Improving the advice. 

<b>But there's a meta level to this ticket ... and it probably needs to be solved first ...</b>

Over time, I've shied away from trying to improve the official Reagent docs because the existing tool chain just doesn't work on Windows (where my clients and I live).  I tried putting docs in the reagent Wiki but noone ever found them, so in the end I put my docs into the re-frame Wiki because there they got some exposure.  But that's clearly not the right place.

So ... 

We need to make the Reagent docs better in many ways - add FAQs etc - but they are currently immutable and incomplete.  But, in my experience, that can't happen until the tool-chain used is improved.  I have no idea how that should happen.  You thoughts?  Abandon the existing custom tool-chain and use something more crossplatform like jekyll?  Move to something hosted like Readme.io?  Other options?

Recent exhibits:  #206, #147,  #191,  #101, etc

escherize:
https://github.com/tripit/slate Is a good candidate. Slate has pretty good deep-linking and looks.

Although its actually made for documenting APIs, there are people using it for libraries/frameworks.

Some examples:
http://developer.aircall.io/#examples <- looks good without language selection
http://appium.io/slate/en/master/?javascript#automating-hybrid-ios-apps

arichiardi:
[...]

Issue #166, opened 2015-08-02
Document reaction, make-reaction, and run!


danielcompton:
There are no docstrings for `reagent.ratom/reaction`, or `make-reaction`. It would be really helpful to explain:
- [ ] The parameters to `make-reaction`
- [ ] How to use `reaction`
- [ ] How `reaction` differs from `run!` and when to use each.

Relates to #116.

JulianLeviston:
The function make-reaction, and its macro reaction are used to create a Reaction, which is a type that belongs to a number of protocols such as IWatchable, IAtom, IReactiveAtom, IDeref, IReset, ISwap, IRunnable, etc. which make it atom-like: ie it can be watched, derefed, reset, swapped on, and additionally, tracks its derefs, behave reactively, and so on.

Reactions are what give r/atom, r/cursor, and function r/cursor and r/wrap their power.

make-reaction takes one argument, f, and an optional options map. f is the function that the reaction you want to make wraps. the options hash can contain an auto-run boolean (if it should automatically run its function on change, I think), on-set and on-dispose which are run when the reaction is set and unset from the DOM repectively. I'm not sure what derefed is. I'm pretty sure someone else can easily fill this in though.

metasoarous:
Wow... is this still outstanding? I've found the re-frame documentation is generally great for more in depth Reagent functionality, but it still surprises me that Reagent doesn't have more documentation about all of it's wonderful goodies. I'm generally pretty informed about ratoms and reactions, but I'd really like to know more about make-reaction in particular. I haven't seen anything about it on the re-frame documentation.

piotr-yuxuan:
Hello :-)

Just a reminder about this. I come from [`re-frame` documentation](https://github.com/Day8/re-frame/blob/master/docs/Subscribing-To-External-Data.md#some-code) and I stumble upon argument `on-dispose`. There are no docs anywhere about these three `on-set`, `auto-run` and `on-dispose` and it may be good to have guideline on how to use it (or perhaps it's fine to remove them from the public API).

Names can be misleading: for instance, a previous comment on this issue states

> the options hash can contain an auto-run boolean

but without documentation, I could also rely on [Clojure naming convention](https://github.com/bbatsov/clojure-style-guide) to say it's not a boolean (would be `auto-run?`) but a function which run each time the reaction is called / changed.

vemv:
Could the notion of a reaction itself be described more clearly?

[...]

PR #343, opened 2018-02-21
Draft user manual.

jmlsf:
This is a preliminary draft of a "React User Manual", which does the following:

1. Gathers all of the information strewn about the release notes
2. Incorporates quite a bit of the information in the /docs directory (I hope that's okay)
3. Adds a section on React interop
4. Attempts to deal with #249 #166 and to a lesser extent #209 and #302

There are quite a few "TODO"s in this documentation where I just don't know the answer and it probably needs another pass through for spelling, but this should get the ball rolling.  Comments welcome.

A few comments on the technology discussion of #209: it seems like the two goals of a different documentation tool are (1) sidebar contents like in gistbook, asciidoctor, or slate and (2) live examples.  My only real comment is that it seems undeniable that the previous format was not really working (even the recent release notes don't use the custom format of the 0.6 and earlier release notes), and I would vote for some kind of markdown-based format.  Perhaps a modern tool + a post processor is the way to go, as people can just edit the markdown and forget about the live examples.  At any rate, I'd rather just get the content in a usable state, so if you can fill in some of the holes, please do.


docs/ReactUserGuide.md
```diff
@@ -0,0 +1,1052 @@
...
+
+```clojure
+[:div>p>b "Nested Element"]
+```
+
+## Rendering Hiccup
+
+The primary entrypoint to the reagent library is `reagent.core/render`.
+
+```
```
gadfly361: add `clojure` after the backticks

docs/ReactUserGuide.md
```diff
@@ -0,0 +1,1052 @@
...
+
+So the rookie mistake is to forget to put in the `[a b c]` parameters on the inner render function.
+
+Remember, `outer` is called once _per component instance_. Each time, the parameters to `outer` will hold the initial parameter values. The renderer on the other hand, will be called by Reagent many times and, each time, potentially with alternative parameter values, but unless you repeat the parameters on the renderer it will close over those initial values in `outer`. As a result, the component renderer will stubbornly only ever render the original parameter values, not the updated ones, which can be baffling for a beginner.
+
+## How Reagent Knows the Difference Between Form-1 and Form-2
+
+When Reagent is turning your Hiccup into React components, (see below for a fuller discussion of the rendering process), it will first invoke the form-1 or form-2 function specified. For example:
+
+```
```
gadfly361: add clojure after the backticks

mike-thompson-day8:
@jmlsf  I greatly appreciate the effort. I know how long docs take. 

Within `/docs` we have a README which acts as something of Table of Contents, and then one doc for each "chapter".    I'd would prefer to continue with that process because it gives us incremental improvements.  (Rather than the big bang approach, of which I'm suspicious). 

I would suggest that we don't need to rewrite or replace the existing docs, but rather supplement them.  I'm definitely keen to focus on the words and worry about fancy rollup later.  

For example, as a first step, I'd be keen to see your "Using Hiccup to Describe HTML" section taken out into a separate document and put into a document called `UsingHiccupToDescribeHTML.md` and added to the README ahead of the existing docs.  It contains good information. 

Then, as a next step, let's identify the further documents we should add one at a time.  Let's do a PR for each one, and get them in incrementally. 

When we have a directory full of all the right material, one document representing each subject, then let's think about one rolled up document.  The incremental approach works best.  Let's not rewrite anything we already have.

jmlsf:
I'm fine with this.  One comment about re-writing: I would suggest that we need to distinguish between tutorial / didactic type documents and reference material.  Reagent, although it describes itself as a "simple" interface to React, is actually really complicated and subtle, and one of its biggest flaws is that there is no reference material.  So to the extent I edited down some of the existing material, it was to that end.

mike-thompson-day8:
Both kinds of docs (tutorial and reference) are useful.  I've tended to be focused on the tutorial kind. I've assumed that your PR was oriented that way too.

mike-thompson-day8:
Many Thanks!

pesterhazy:
Just saw this, great work @jmlsf!

PR #347, opened 2018-02-27
A few added details on ratom updates.

jmlsf:


mike-thompson-day8:
Can you explain more what your thinking is behind this:  
```
3b. If you put your ratom in a form-2 or form-3 component, be sure you are calling that function using \[square brackets\], not \(parenthesis\). 
```

jmlsf:
Okay this is more complicated than I described.  Hopefully the clojurians slack log staging site will come back online because I encountered this problem on #reagent.  Let's just put this edit on hold until I remember what you have to do to hit this behavior.

Deraen:
3b -> If function (component) is called using `(` `)` Reagent doesn't create new component, but instead just places the function's return value into current component. In this case the function closure which should hold the local state doesn't work.

PR #351, opened 2018-03-12
Keep reference to real synthetic input component

Deraen:
To implement proper on-update function for React input wrapper components, reference to the real component is probably needed.

Discussion at https://github.com/madvas/cljs-react-material-ui/issues/17
Example at https://github.com/Deraen/problems/blob/reagent-material-ui-text-field/src/hello_world/core.cljs#L16

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -213,7 +213,7 @@
           synthetic-on-update ($ this :cljsSyntheticOnUpdate)]
       (when (not= rendered-value dom-value)
         (if (fn? synthetic-on-update)
-          (synthetic-on-update input-node-set-value node rendered-value dom-value this)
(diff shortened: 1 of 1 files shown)


Deraen:
I've been looking to finalizing this, but seems like this approach can't work with material-ui 1.0 as `ref` can't be used with new `TextField` which is implemented as Functional Stateless Component. This means it is impossible to get access to the React component.

I'm starting to think it will be near impossible to fix this until React starts supporting async rendering (could be soon).

I'll keep thinking about this and if I find solution, before I'll merge it or release 0.8 I'll add examples for Material-UI, both 0.20 and 1.0-beta and React-bootstrap so we can be sure the approach works for different cases.

Deraen:
Proper solution has been found for v1: https://github.com/reagent-project/reagent/blob/master/examples/material-ui/src/example/core.cljs

PR #354, opened 2018-03-14
Add link to FAQ and tutorials

nbardy:
Not sure if this belong in the readme yet because they're some draft being linked to, but I found these documentations more useful than the API Documentation. I think it's worth pointing to.


mike-thompson-day8:
Thanks

PR #368, opened 2018-04-27
Class fixes

### LINKED ISSUE #367: :class no longer supports keywords (since 0.8.0)
Before Reagent 0.8.0 it was possible to use a keyword as a class (this seems to be facilitated by `convert-prop-value` being called?).

Reagent 0.8.0 breaks this by changing e.g the keyword `:highlight` to the string `":highlight"` (including the `:`, indicating that `name` hasn't been called). In earlier versions of Reagent this would result in the string `"highlight"` and a working class.

Possibly introduced in https://github.com/reagent-project/reagent/pull/323?

Deraen:
Fixes #367

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## Unreleased
+
+**[compare](https://github.com/reagent-project/reagent/compare/v0.8.0...master)**
+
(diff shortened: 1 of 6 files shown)


Deraen:
I modified this to check `named?` before calling `name` and added tests for numbers to validate even non-named types can be used.

PR #380, opened 2018-06-17
Alternative mui input fix

Deraen:
Alternative custom input component fix. This might be more general.


examples/material-ui/src/example/core.cljs
```diff
@@ -7,42 +7,28 @@
...
-                  rtpl/convert-prop-value)]
-    (apply r/create-element mui/TextField props (map r/as-element children))))
+(defn adapt-input-component [component]
+  (fn [props & _]
+    (r/create-class
+      {:getInitialState (fn [] #js {:value (:value props)})
+       :component-will-receive-props
+       (fn [this [_ next-props]]
+         (when (not= (:value next-props) (.-value (.-state this)))
+           (.setState this #js {:value (:value next-props)})))
```
pesterhazy: I guess .forceUpdate is not needed here because the props change?

examples/material-ui/src/example/core.cljs
```diff
@@ -7,42 +7,28 @@
...
-                                                            nil
-
-                                                            :else
-                                                            input-component))
-                  rtpl/convert-prop-value)]
-    (apply r/create-element mui/TextField props (map r/as-element children))))
+(defn adapt-input-component [component]
+  (fn [props & _]
+    (r/create-class
+      {:getInitialState (fn [] #js {:value (:value props)})
```
pesterhazy: AFAIK state doesn't have to be a map so you could just use `(:value props)` here

examples/material-ui/src/example/core.cljs
Deraen: Yes. And I just replace `forceUpdate` with `shouldComponentUpdate`.

pesterhazy: Seems cleaner yes! Is `(constantly true)` the default for shouldComponentUpdate?

pesterhazy: According to the docs, yes: https://stackoverflow.com/questions/33466651/what-is-the-default-implementation-for-shouldcomponentupdate-lifecycle-method

Deraen: Yes, it is React default. Reagent default is to only check props.

examples/material-ui/src/example/core.cljs
```diff
@@ -7,42 +7,31 @@
...
+         true)
+       :reagent-render
+       (fn [props & children]
+         (this-as this
+           (let [props (-> props
+                           (cond-> (:on-change props)
+                             (assoc :on-change (fn [e]
+                                                 (.setState this #js {:value (.. e -target -value)})
+                                                 ((:on-change props) e))))
+                           (cond-> (:value props)
```
pesterhazy: Maybe only apply these two `assoc`s only if _both_  on-change and value are set?

Deraen:
As a test, replaced the whole input workaround in reagent.impl.template with this solution. Testing is required to see if this is usable solution. Is the performance similar, does this work in IE... etc.

Deraen:
Okay, this won't work because the existing selection logic is required for cases where input value e.g. has transformation, like upper case.

jmlsf:
This is actually a react problem, not a reagent problem.  React doesn't try to maintain cursor positioning when there is a transformation because transformations that alter the number of characters don't have a general solution to the question of where to place the cursor.  See https://github.com/facebook/react/issues/9894.

I confirmed in a simple example that this is the case with plain react:

```js
class App extends Component {
  constructor(props) {
    super(props);
    this.state = {
      value1: '',
      value2: ''
    };
  }
  render() {
    return (
      <div className="App">
        {/* Cursor is maintained properly: */}
        <input
          type="text"
          onChange={(e) => this.setState({value1: e.target.value})}
          value={this.state.value1} />
        {/* Cursor jumps to the end: */}
        <input
          type="text"
          onChange={(e) => this.setState({value2: e.target.value.toUpperCase()})}
          value={this.state.value2} />
      </div>
    );
  }
}
```

(Also I don't think it is foolish to allow transformations.  It allows things like phone number and credit card formatting.  But you do have to manage the cursor manually if you're going to do that.)

PR #381, opened 2018-06-18
Alternative custom input fix

Deraen:
While this implementation can't be used to replace existing logic in Reagent (#380), and it has some trade-offs compared to the other Material-UI solution, it might make sense to provide this function for some use cases.

### DIFF
--- a/examples/material-ui/src/example/core.cljs
+++ b/examples/material-ui/src/example/core.cljs
@@ -7,42 +7,42 @@
 (def mui-theme-provider (r/adapt-react-class mui/MuiThemeProvider))
 (def menu-item (r/adapt-react-class mui/MenuItem))
 
-(def ^:private input-component
-  (r/reactify-component
-    (fn [props]
-      [:input (-> props
-                  (assoc :ref (:inputRef props))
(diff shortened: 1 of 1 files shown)


pesterhazy:
Just tried this, works great! It's pretty easy to test with regular input fields as well:

```
(defn input-ui [{:keys [value onChange] :as props}]
  [:input {:value value
           :on-change (fn [e] (onChange (-> e .-target .-value)))}])

(def input-ui-plain (r/reactify-component input-ui))

(def input-ui-plain* (adapt-input-component input-ui-plain))

(defn handle-change [v]
  (swap! !state assoc :text v))

(def handle-change* (gfun/debounce handle-change 1000))

(defn test-ui []
  [input-ui-plain* {:value (:text @!state)
                    :on-change handle-change*}])
```

One problem that I've noticed when doing that is with the synthetic input events. React wants you to use the synthetic input event right away (in the event handler); if you don't, it gives you the "This synthetic event is reused for performance reasons" warning.

For me the solution is to extract the value from the input event in the `input-ui` wrapper, turning the event (which apparently is a  mutable object!) into a value which can be passed around without problems.

PR #391, opened 2018-08-01
Docs: Add special notation for id and classes

grav:
I added docs for the special notation of ids and classes (`#id` and `.class`). 

More or less stolen from here: 
https://www.reddit.com/r/Clojure/comments/75d054/hiccupreagent_syntax_help/do5o0qh/


gadfly361:
@grav Thanks for this!

PR #395, opened 2018-08-26
material demo with reframe and routing

headwinds:
screen shot 2018-08-26 at 10 33 56 am

Inspired by the [material ui](https://material-ui.com/) site which has a really nice look and feel, I thought would be also nice to start to clean up the demo and model it after its JS source but built all in ClojureScript!

This PR represents a start at this new design goal and begins to outline the skeleton but the demos are not fully fleshed out yet.


headwinds:
I wanted to a [demo link](https://intense-eyrie-34041.herokuapp.com/) and took this opportunity to wrestle with heroku and finally had a break through so I've added some notes on deploying

Deraen:
The purpose of material-ui example is to only demonstrate the JS interop and few workarounds for material-ui inputs. This change would introduce lots of complexity that is not related to those purposes.

I have no interest in maintaining re-frame, heroku, compojure or such examples in Reagent.

Such example is better maintained as separate repository.

headwinds:
@Deraen ok that makes sense - I'll move it into its own repo - would you be open to cross promote it by updating your readme? I really like what you did by showing how to add material ui and would like others to benefit from learning how reagent works with it.

headwinds:
I've moved it this [repo](https://github.com/headwinds/reagent-reframe-material-ui) - I'll look at the open Reagent issues and see if there is a better way for me to contribute ;-D

Issue #71, opened 2014-11-26
Reagent logo


Frozenlock:
I saw @gadfly361's logo for the Reagent Cookbook and thought Reagent could use one as well.
He was kind enough to send me his SVG file.

I cleaned it up a little and submit it here.

What do you guys think?
Good logo?

https://dl.dropboxusercontent.com/u/40737386/reagent-logo.svg
![reagent-logo](https://cloud.githubusercontent.com/assets/1152041/5208084/a9307afe-757e-11e4-9b05-74324010be1f.png)

https://dl.dropboxusercontent.com/u/40737386/reagent-logo-only.svg
![reagent-logo-only](https://cloud.githubusercontent.com/assets/1152041/5208105/e2b849fa-757e-11e4-9c70-93c57789c504.png)

Frozenlock:
I'll close this for now, as it seems we have a *de facto* logo.

PR #400, opened 2018-09-11
Add logo

codingthat:
This seems to be the "de facto" logo mentioned in #71 being that it's in the cookbook and template, so I thought it should probably go here too.


gadfly361:
@codingthat Good idea, thanks for the PR!

Deraen:
@gadfly361 Do you still have svg version of the logo? I'd like to add that to the repo also.

gadfly361:
@Deraen I sent you an email with the most recent version I could find (but it isn't the dark version that is currently floating around).

PR #412, opened 2018-12-03
Allows for an arbitrary number of props in merge-props

jimberlage:
We've run into a few instances of people trying to use `reagent.core/merge-props` like `clojure.core/merge`, with multiple maps.  For an example, think of something like this:

```clojure
(r/merge-props
  {:style {:display "flex" :flex-direction "flex-column"}}
  (when @loading
    {:disabled true {:style {:background-color "gray"}}})
  (when @demo
    {:class "demo-list"}))
```

This doesn't change the behavior of `merge-props` for existing code.  (Unless people were relying on having `(merge-props {})` throw an error?)

### DIFF
--- a/src/reagent/core.cljs
+++ b/src/reagent/core.cljs
@@ -195,10 +195,12 @@
   (dom/dom-node this))
 
 (defn merge-props
-  "Utility function that merges two maps, handling :class and :style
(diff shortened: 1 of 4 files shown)


Deraen:
Merged with some tuning (fixed `:class` collection handling) and I also added `class-names` utility function.

PR #415, opened 2018-12-08
Fix InteropWithReact.md for "Don’t Use HOCs Inside the render Method"

boxp:
https://reactjs.org/docs/higher-order-components.html#dont-use-hocs-inside-the-render-method


boxp:
Thanks!

PR #430, opened 2019-05-14
Support getSnapshotBeforeUpdate

vnorilo:
- Adds a reagent-style wrapper for getSnapshotBeforeUpdate
- Relays the possible snapshot parameter from react to the user callback


Deraen:
Related docs:

https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate
https://reactjs.org/docs/react-component.html#componentdidupdate

src/reagent/impl/component.cljs
```diff
@@ -192,9 +192,13 @@
     (fn componentWillUpdate [nextprops]
       (this-as c (.call f c c (props-argv c nextprops))))
 
+    :getSnapshotBeforeUpdate
+    (fn getSnapshotBeforeUpdate [_ _]
+      (this-as c (.call f c c)))
```
Deraen: Maybe this fn should take `oldprops` & handle them similar to `componentDidUpdate`?

vnorilo:
I just saw this:
```
Warning: Unsafe legacy lifecycles will not be called for components using new component APIs.

veneer.patcher.node.render_node uses getSnapshotBeforeUpdate() but also contains the following legacy lifecycles:
  componentWillMount
```

Looks like actually using snapshots might interfere with the internal Reagent life cycle. I'll have a look; please do not merge yet!

vnorilo:
See f.e. https://blog.northcoders.com/react-componentwillmount-to-be-deprecated

Looks like the internal lifecycle work done in `componentWillMount`:
`(set! (.-cljsMountOrder c) (batch/next-mount-count))`

Could be moved to `getSnapShotBeforeUpdate` as that is called at a similar time.

vnorilo:
For now, it might be best to use `componentWillMount_UNSAFE` and translate `componentWillMount` internally to that.

vnorilo:
Could `.-cljsMountOrder` be set in `.getInitialState` instead? This would remove the dependency on the deprecated `componentWillMount`. According to https://reactjs.org/docs/react-component.html#mounting it seems like so.

vnorilo:
Closing this PR because it appears support for the new lifecycle methods is best done in an overall upgrade to react 17 semantics.

PR #437, opened 2019-07-29
Enhance Docs w/ link to reagent component blog

tkjone:
`what the reagent component` is a blog post that gets into the nitty gritty details of what it means when the docs say "underneath they are all just React components".  

If there is anything missing from a technical perspective, I am happy to correct 😄

### DIFF
--- a/doc/CreatingReagentComponents.md
+++ b/doc/CreatingReagentComponents.md
@@ -231,7 +231,7 @@ Leaving out the `:display-name` entry.  If you leave it out, Reagent and React h
 
 Above I used the terms `Form-1`, `Form-2` and `Form-3`, but there's actually only one kind of component. It is just that there's **3 different ways to create a component**. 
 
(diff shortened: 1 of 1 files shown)


Deraen:
Thanks! Very nice post.

> For those who want to know why the Reagent team chose to modify the pseudoclassical instantiation pattern I do not really have an answer. At the end of the day, they do more or less the same things without any significant downsides.

Me neither :) But I wasn't even aware of the differences, so I guess this will be something I'll look into, especially if there is something which could improve the performance of Reagent.

Note, in general I'd prefer having the information here in the Reagent docs instead of blog posts. It will be easier to update local docs.

tkjone:
> Note, in general I'd prefer having the information here in the Reagent docs instead of blog posts

Noted 😄 I was thinking about posting this in the docs, but it is...substantial lol.  Thanks for the merge 👍

PR #442, opened 2019-08-12
Remove deprecated dependency cljsjs/react-dom-server

superstructor:
It is [deprecated](https://github.com/cljsjs/packages/blob/master/react/build.boot#L68) and included in the react-dom package.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -8,8 +8,7 @@
                  ;; Without direct react dependency, other packages,
                  ;; like react-leaflet might have closer dependency to a other version.
                  [cljsjs/react "16.8.6-0"]
-                 [cljsjs/react-dom "16.8.6-0"]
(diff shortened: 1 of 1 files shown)


Deraen:
I'll have to test this a bit.

I deliberately left the dependency on the react-dom-server when I deprecated it. If we just remove the dependency, user might easily get the old version of react-dom-server package from some transitive dependency, which could overwrite the newer files in new react-dom package.

Currently react-dom-server package is created for all new React releases, but it is empty.

superstructor:
Apologies I didn't actually test this. I just assumed based on the [code comments](https://github.com/cljsjs/packages/blob/master/react/build.boot#L68). Thus your caution may be well founded.

superstructor:
I think this can be closed as per previous discussion plus branch has conflicts. You can revisit in a new change if you think its appropriate.

PR #443, opened 2019-08-13
React 16.9

Deraen:
- Adds support for using `UNSAFE_` lifecycle methods.
- Migrates Reagent implementation away from `componentWillMount` to `componentDidMount` to avoid deprecation warnings.
- Separated `getInitialState` and `constructor` methods, they work differently
- Tested & improved support for `getDerivedState` methods and `getSnapshotBeforeUpdate`

I'm also interested in trying out if we could generate functional components in some cases, but I guess that will be a separate, bigger, PR.

TODO: I'll test this in few work projects to ensure everything is still working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,13 @@
 
 **[compare](https://github.com/reagent-project/reagent/compare/v0.8.1...master)**
 
+- Default to React 16.9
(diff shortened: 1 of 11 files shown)


Deraen:
I've pushed this to Clojars as `[reagent "0.9.0-SNAPSHOT"]`

PR #450, opened 2019-09-10
Optimize dash-to-prop-name & capitalize fns

roman01la:
This PR optimizes string manipulation code in `reagent.impl.util/dash-to-prop-name` and `reagent.impl.util/capitalize` which results in ~34% speedup when interpreting Hiccup, according to synthetic benchmark added here.

I suspect similar optimizations could be done in other parts of code base, but this can be done in separate PRs.


src/reagent/impl/util.cljs
```diff
@@ -24,18 +24,29 @@
...
     (let [name-str (name dashed)
-          [start & parts] (string/split name-str #"-")]
-      (if (dont-camel-case start)
-        name-str
-        (apply str start (map capitalize parts))))))
+          [start & parts] (.split name-str #"-")]
+      (cond
+        (contains? dont-camel-case start) name-str
+
+        (not (nil? parts))
```
Deraen: I think this could be just `parts`?

src/reagent/impl/util.cljs
```diff
@@ -24,18 +24,29 @@
...
+        (-> parts
+            (array-reduce
+              (fn [a p]
+                (.push a (capitalize p))
+                a)
+              #js [])
+            (.join "")
+            (->> (str start)))
```
Deraen: `start` could be moved to reduce initial value array? Not sure how complex `str` implementation is, but this could be a bit faster.

roman01la: You are right, good catch

src/reagent/impl/util.cljs
roman01la: Leaving just `parts` is going to generate truthy check around it, simpler form would be `(some? parts)`. I'd prefer to explicitly coerce `parts` to `boolean` instead of type hinting it with `^boolean`, so checked if can be eliminated.

Deraen:
So, I did some testing with this:

- Currently this doesn't work as destructuring will return `parts` as Seq, not Array
- `array-reduce` is private in cljs.core
- I can't see any performance different with Firefox or Chrome here, 5000ms with the old function and the new in both browsers.
- Test case contains lots of hiccup tags with classes, but this change doesn't affect tag name parsing, which might explain why I don't see change.

I tested version like this, where destructuring is not used so split result is kept as Array:

```cljs
(defn dash-to-prop-name [dashed]
  (if (string? dashed)
    dashed
    (let [name-str (name dashed)
          parts (.split name-str #"-")
          start (aget parts 0)]
      (cond
        (contains? dont-camel-case start) name-str

        (> (.-length parts) 1)
        (-> (array-reduce
              parts
              (fn [a p]
                (.push a (capitalize p))
                a)
              #js [start]
              1)
            (.join ""))

        :else start))))
```

roman01la:
I think we can use normal `reduce` here and it will fall into `array-reduce` when applied to `js/Array`. Perf difference should be noticeable for Hiccup with lots of attributes and inline styles.

Deraen:
Yes, `reduce` is probably fine. Though it doesn't directly call `arrray-reduce`. When destructuring is used to get `start` and `parts`, `parts` will be `ArrayChunk` which has it's own `IReduce` implementation using `array-reduce` and offset.

I'll look more into this after getting 0.9 finally out.

Probably benchmarking the `dash-to-prop-name` alone, preferably with Benchmark.js, should provide more information about this change.

roman01la:
Pushed an update to avoid implicit conversion to a seq when destructuring js/array

roman01la:
@Deraen Is there anything blocking this PR from getting merged?

Deraen:
I think I was unsure of the performance improvements last I tried to benchmark this.

> Probably benchmarking the dash-to-prop-name alone, preferably with Benchmark.js, should provide more information about this change.

PR #451, opened 2019-09-20
Added Dependency to help with compilers

eskemojoe007:
This is one of the few working examples of Material-UI with reagent.  When building for prod, it is needed to include `[cljsjs.material-ui]`, so I added that to leave a bread crumb for new comers.


Deraen:
Thanks for the PR, but this is not correct. When using the cljsjs package, `cljsjs.material-ui` is just old name for `"@material-ui/core"` which is already required. The old name is provided just for compatibility with old projects and should not be used.

I've tested this example with advanced compilation and it works. If you have problems, please provide more detailed description of the problem.

Issue #449, opened 2019-09-10
Reagent 0.9.0-rc reports


Deraen:
Before pushing out final 0.9.0 release, I'd like to hear reports if this works in existing projects, or if there are regressions. Particularly I'm afraid about mount order change, related to lifecycle method changes: https://github.com/reagent-project/reagent/commit/35ff5d33dd9dda6ac4103bb1bcb88580bdd12ddd

https://clojars.org/reagent/versions/0.9.0-rc1

superstructor:
I'll test this with all the re-frame example apps and re-frame template this evening. Thanks :+1:

superstructor:
As a [first test](https://github.com/Day8/re-frame/commit/7b0f36ce18a9fbe1ba4e6b7fe2d11fbe6132ae2c) re-frame's [tests pass](https://circleci.com/gh/Day8/re-frame/1512) and the two example apps in the same repo (simple and todomvc) work as expected.

PrestanceDesign:
I tested on this two React Native templates and that works :
* https://github.com/thheller/reagent-expo
* https://github.com/seantempesta/expo-cljs-template


I'll try on Re-natal tomorrow : https://github.com/seantempesta/expo-cljs-template

superstructor:
I have released re-frame `0.11.0-rc1` that depends on reagent `0.9.0-rc1` for those who want to use it for testing re-frame apps.

oceangroove:
Hi, I happen to upgrade from 0.8.1 to 0.9.0-rc1 whilst trying to fix an optimisation bug in my clsls/material-ui app.  The problem is with MuiThemeProvider which causes a failure with :optimizations :advanced. but works with :none and :simple

After the reagent version change the error now occurs in all modes.

After the change of reagent version my tested demo and dev versions now produce the following :

    Uncaught Error: No protocol method IKVReduce.-kv-reduce defined for type object: [object 
   Object]

This is a bug in my code but the behaviour has changed with the new version.

logCapturedError | @ | react-dom.inc.js:19952
-- | -- | --
  | logError | @ | react-dom.inc.js:19988
  | update.callback | @ | react-dom.inc.js:21067
  | callCallback | @ | react-dom.inc.js:13067
  | commitUpdateEffects | @ | react-dom.inc.js:13106
  | commitUpdateQueue | @ | react-dom.inc.js:13097
  | commitLifeCycles | @ | react-dom.inc.js:20236
[...]

PR #461, opened 2019-10-25
Fix #460: Add type hints for extern inference

### LINKED ISSUE #460: 0.9.0-rc1 TypeError: Cannot read property '$argv$' of undefined
This happening when compile with `:advanced`. It works during developing `clj -A:fig:build`, because no `:advanced` compilation.

Reagent version:
`reagent {:mvn/version "0.9.0-rc1" :exclusions [cljsjs/react cljsjs/react-dom]}`

It works with:
`reagent {:mvn/version "0.8.1" :exclusions [cljsjs/react cljsjs/react-dom]}`

image

Error:
```
git clone -b doc git@github.com:kwladyka/form-validator-cljs.git doc
cd doc
git checkout a81d4bc574cbab68e67a791db118e44f373e3217
# add `:pseudo-names true` to `dev.cljs.edn`
npm i
npx webpack --mode=production
clojure -A:fig:min
cp resources/public/index.html target/public/
cp -r resources/public/css target/public/
```

Then open `target/public/index.html` to see the issue.

Temporary solution: downgrade Reagent to `0.8.1`

Deraen:
TODO: Run shadow-cljs or figwheel-main tests with advanced compilation because this is probably currently missing some hints.

### DIFF
--- a/src/reagent/impl/batching.cljs
+++ b/src/reagent/impl/batching.cljs
@@ -24,7 +24,7 @@
           fake-raf))))
 
 (defn compare-mount-order
-  [c1 c2]
+  [^clj c1 ^clj c2]
   ;; Mount order is now set in DidMount method.  I.e. the
   ;; top-most component is mounted last and gets largest
   ;; number. This is reverse compared to WillMount where method
@@ -49,7 +49,7 @@
   (dotimes [i (alength fs)]
     ((aget fs i))))
 
-(defn enqueue [queue fs f]
+(defn enqueue [^clj queue fs f]
   (assert-some f "Enqueued function")
   (.push fs f)
   (.schedule queue))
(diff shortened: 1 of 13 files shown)


src/reagent/impl/batching.cljs
```diff
@@ -37,7 +37,7 @@
   ;; are rendered before children
   (.sort a compare-mount-order)
   (dotimes [i (alength a)]
-    (let [c (aget a i)]
+    (let [^js/React.Component c (aget a i)]
```
lins05: I guess this is the trick that fixes the advanced compile bug I see in https://github.com/reagent-project/reagent/issues/449#issuecomment-554625425 ? I wonder how could non-shadow-cljs tools work without this?

Deraen: https://clojurescript.org/guides/externs
https://github.com/cljsjs/packages/blob/master/react/resources/cljsjs/react/common/react.ext.js#L149
https://shadow-cljs.github.io/docs/UsersGuide.html#_why_not_use_cljsjs

lins05: I see, so other tools would use the externs for react because it's bundled in cljsjs, while shadow-cljs can't make use of it. Thanks for the links!

PR #468, opened 2019-12-16
Use component constructor to keep track of mount order

### LINKED ISSUE #462: [0.9.0-rc2] child subscriptions are still run on unmount
This isn't confirmed yet, but we recently had to revert to 0.8.1 after trying 0.9.0-rc2 due to a bug where subscriptions to re-frame events were being run when the change to the data should have caused the children to be unmounted.

I hope to have time soon to create a minimal repo, but in the mean time I wanted to record it here in case others were experiencing the same issue and/or wanted to investigate.

Deraen:
Parent components need to have smaller mount-order numbers than their
children, or else this will break when only the children are rerendered
and get new number.

Fixes #462

### DIFF
--- a/src/reagent/impl/batching.cljs
+++ b/src/reagent/impl/batching.cljs
@@ -29,8 +29,8 @@
   ;; top-most component is mounted last and gets largest
   ;; number. This is reverse compared to WillMount where method
   ;; for top component gets called first.
-  (- (.-cljsMountOrder c2)
-     (.-cljsMountOrder c1)))
+  (- (.-cljsMountOrder c1)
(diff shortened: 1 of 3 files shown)


src/reagent/impl/batching.cljs
```diff
@@ -29,8 +29,8 @@
   ;; top-most component is mounted last and gets largest
```
MrMikeRodriguez: So comment is wrong now?

mrrodriguez: So comment is wrong now?

PR #469, opened 2019-12-17
Reagent Next

Deraen:
Starting work on what I think will become Reagent 1.0.

I have some interesting plans regarding allowing more configuration, but I'm starting with cleaning the API. Changelog will mention current changes.

TODO:

- [ ] Remove dependency on Cljsjs and instead document on how users should provide React themselves
- [ ] Ensure Reagent works without externs in all cases (Node module processing especially problematic?)

### DIFF
--- a/src/reagent/impl/component.cljs
+++ b/src/reagent/impl/component.cljs
@@ -431,3 +431,64 @@
   (if (react-class? comp)
     comp
     (as-class comp)))
+
+(defonce fun-component-state #js {})
+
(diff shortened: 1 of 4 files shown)


Deraen:
Latest commits test an idea for creating functional components from Reagent hiccup. Currently this is only done for `[:< function ...]` markup, but hopefully this can be the default later.

The code runs the render code in a Reaction, like stateful components, but uses State hook to force re-render when dependent RAtoms update.

PR #472, opened 2020-01-03
Include path in assert error for cursors

dazld:
Hopefully a fairly uncontroversial PR! 

Including path can help track down which particular cursor is misbehaving, ie in a React Native project, I have the following error message:

image

This isn't particularly helpful, but when the path is included it becomes a lot easier to find which particular cursor is responsible for the assert failure.

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -321,7 +321,9 @@
               (and (ifn? src)
                    (not (vector? src))))
           (str "src must be a reactive atom or a function, not "
-               (pr-str src)))
+               (pr-str src)
+               " while attempting to get path: "
+               (pr-str path))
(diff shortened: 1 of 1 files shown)


src/reagent/ratom.cljs
```diff
@@ -321,7 +321,9 @@
               (and (ifn? src)
                    (not (vector? src))))
           (str "src must be a reactive atom or a function, not "
-               (pr-str src)))
+               (pr-str src)
+               "while attempting to get path: "
```
Deraen: I think this is missing space on the beginning?

dazld: good spot, will fix.

dazld:
@Deraen fixed your spot - I could expand the PR to look at other `assert` calls too if that's helpful.

PR #474, opened 2020-01-04
Fix example code

dosbol:


mike-thompson-day8:
Hmm. The existing code is valid.  What do you believe needs fixing?

dosbol:
Sorry, I didn't know about optional name in fn.
Thanks.

PR #477, opened 2020-02-05
Test creating functional components

Deraen:
Current status: nearly all test cases are working with the new implementation, done some testing with real work projects and seems to be working.

Next steps:
- Create snapshot release for testing
- Find a way to compare performance to Class components

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@
 
 ### Features and changes
 
+- **Option to render Reagent components as React functional components instead of
+class components**
+    - To ensure backwards compatibility by default, Reagent works as previously and
+    by default creates class components.
+    - New Compiler object can be created and passed to functions to control
+    how Reagent converts Hiccup-style markup to React components and classes:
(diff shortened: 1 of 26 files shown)


Deraen:
Creating functional components is now optional, and can be enabled by providing options to render call:

`(rdom/render [...] {:functional-reag-elements? true})`

This will affect the tree of components created directly from given component. Using `as-element` or `create-class` will require providing the options again.

This is just first version of testing the idea of providing options to control Reagent, but it looks promising. I'm still unsure if having to provide options for `as-element` and such is a problem, and we could just encourage users to create their own namespace where they wrap the functions with their options.

Might be also possible to use Context or maybe even dynamic var, to set the default options for calls without options.

Another problem with this version is, that several of the caches need to be per options, but currently are global. This will probably be solved by creating Compiler type, where Compiler instance contains the state for caches, this compiler is created using the options and then passed to functions.

lucywang000:
@Deraen I did some tests and found that the child components is still rendered as class components. Addressed that in pr https://github.com/reagent-project/reagent/pull/493

PR #478, opened 2020-02-06
Use the correct alias in example

stoyle:
Either that or change the `:as r` to `:as rdom`?


Deraen:
I've fixed the example to require `reagent.dom` as `rdom`. Note that the docs in the repository are for the next release of Reagent, where `render` function is no longer included in the core ns.

PR #479, opened 2020-03-06
Fix merge-props without classes

achikin:
This PR fixes a minor issue with `merge-props`. When called with 2 or more arguments without a `:class` key `merge-props` returns a result with `:class nil`. `semantic-ui-react` interprets the presence of `:class` key with value `nil`  as "remove all classes from the component".

This fix does not add `:class` key at all if no classes are present in the arguments to `merge-props`.

[Test run log](https://gist.github.com/achikin/e01c6106ec824d9d2faa9024d67cd154)

### DIFF
--- a/src/reagent/impl/util.cljs
+++ b/src/reagent/impl/util.cljs
@@ -147,7 +147,9 @@
            rst)))
 
 (defn- merge-class [p1 p2]
-  (assoc p2 :class (class-names (:class p1) (:class p2))))
+  (if-let [names (class-names (:class p1) (:class p2))]
+    (assoc p2 :class names)
+    p2))
 
 (defn- merge-style [p1 p2]
   (let [style (when-let [s1 (:style p1)]
--- a/src/reagent/impl/util.cljs
+++ b/src/reagent/impl/util.cljs
@@ -147,8 +147,8 @@
            rst)))
 
 (defn- merge-class [p1 p2]
-  (if-let [names (class-names (:class p1) (:class p2))]
-    (assoc p2 :class names)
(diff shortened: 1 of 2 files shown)


achikin:
I did not find the "Unreleased" section in [CHANGELOG.md](https://github.com/reagent-project/reagent/blob/master/CHANGELOG.md) to add a change there as per [contributing guidelines](https://github.com/reagent-project/reagent/blob/master/CONTRIBUTING.md). Could someone please help me with that?

achikin:
Any updates?

src/reagent/impl/util.cljs
```diff
@@ -147,7 +147,9 @@
            rst)))
 
 (defn- merge-class [p1 p2]
-  (assoc p2 :class (class-names (:class p1) (:class p2))))
+  (if-let [names (class-names (:class p1) (:class p2))]
+    (assoc p2 :class names)
+    p2))
```
p-himik: What if someone does want `:class nil` to be propagated?

achikin: Good catch! I propose the following solution: if any of class keys exist - then add `:class` key to the result.

achikin: Btw. I came across this issue while updating Reagent from `0.8.1` to `0.9.0`, so before 0.9 `:class nil` was omitted from the result.

Deraen:
Looks good. I'll probably look into this more closely during the weekend.

Between 0.8.1 and 0.9 at least this change could have affected this: https://github.com/reagent-project/reagent/pull/412/files

PR #492, opened 2020-04-22
Polish key-from-vec

zelark:
Don't know if it's acceptable change, but I think now `key-from-vec` fn looks a bit better. Also I replaced `(nth v idx nil)` with `(get v idx)`. The latter doesn't throw IndexOutOfBounds exception, and at least not slower than the former.

```
(js/console.log "nth" (time (dotimes [x 50000000] (nth [1 2 3] 3 nil))))
;; => "Elapsed time: 3583.530000 msecs"

(js/console.log "get" (time (dotimes [x 50000000] (get [2 3 4] 3))))
;; => "Elapsed time: 2635.350000 msecs"
```

### DIFF
--- a/src/reagent/impl/template.cljs
+++ b/src/reagent/impl/template.cljs
@@ -284,14 +284,10 @@
   (when (map? x)
     (try-get-key x)))
 
-(defn key-from-vec [v]
-  (if-some [k (-> (meta v) get-key)]
-    k
-    (or (-> v (nth 1 nil) get-key)
-        ;; :> is a special case because properties map is the first
-        ;; element of the vector.
-        (if (= :> (nth v 0 nil))
-          (get-key (nth v 2 nil))))))
+(def key-from-vec
+  (some-fn #(get-key (meta %))
+           #(get-key (get % 1))
+           #(when (= (get % 0) :>) (get-key (get % 2)))))
 
 (defn reag-element [tag v]
   (let [c (comp/as-class tag)


Deraen:
If you are looking at the performance, it is probably good to check the whole function. Single fn with `or` instead of `some-fn` would probably be faster?

zelark:
Let me check it.

zelark:
Here we go, run it 3 times
```clojure
(defn try-get-key [x]
  ;; try catch to avoid clojurescript peculiarity with
  ;; sorted-maps with keys that are numbers
  (try (get x :key)
       (catch :default e)))

(defn get-key [x]
  (when (map? x)
    (try-get-key x)))

(defn key-from-vec-1 [v]
  (if-some [k (-> (meta v) get-key)]
    k
    (or (-> v (nth 1 nil) get-key)
        ;; :> is a special case because properties map is the first
        ;; element of the vector.
        (if (= :> (nth v 0 nil))
          (get-key (nth v 2 nil))))))

(def key-from-vec-2
  (some-fn #(get-key (meta %))
           #(get-key (get % 1))
           #(when (= (get % 0) :>) (get-key (get % 2)))))

(js/console.log (time (dotimes [x 50000000] (key-from-vec-1 [:> "div" {:key 1}]))))
;; => "Elapsed time: 31631.195000 msecs"
;; => "Elapsed time: 29565.440000 msecs"
;; => "Elapsed time: 28984.810000 msecs"

(js/console.log (time (dotimes [x 50000000] (key-from-vec-2 [:> "div" {:key 2}]))))
;; => "Elapsed time: 26881.605000 msecs"
;; => "Elapsed time: 24992.270000 msecs"
;; => "Elapsed time: 25282.350000 msecs"
```

zelark:
In `meta` case the results are as following:
```
;; key-from-vec-1
;; => "Elapsed time: 14995.210000 msecs"
;; => "Elapsed time: 16679.270000 msecs"
;; => "Elapsed time: 12559.060000 msecs"

;; key-from-vec-2
;; => "Elapsed time: 14797.765000 msecs"
;; => "Elapsed time: 15881.220000 msecs"
;; => "Elapsed time: 12358.880000 msecs"
```

Deraen:
I did some testing with cleaned defn with or version, and some-fn and there wasn't much performance difference. I might refactor the places where this is called, so that this doesn't need to check for `:>` or other shortcuts.

zelark:
@danielcompton thanks for responding and some extra testing. Actually I also did some extra tests with advanced compilation mode and simple-benchmark and didn't notice some significant difference between the original variant and mine. But my initial point was to improve a bit readability. ^_^

However, during that testing I realised that for the first case we can easily skip `get-key` because 'meta' is always regular map. Thus we can just write something `(-> v meta :key)`, and it will give great boost.

Deraen:
Thanks for the PR. I've reimplemented this a bit differently, as I had moved the function to another file.

zelark:
Awesome!

PR #493, opened 2020-04-22
Fix child functional components

lucywang000:


Deraen:
Do you have some tests for this?

I've changed the implementation a lot and merged the branch to master. Please test with that and provide a test case if this still is a problem.

lucywang000:
works on 1.0.0-snapshot, cool!

PR #498, opened 2020-05-01
Fix error-boundary example

davidjameshumphreys:
Re-balance parens.
Read the error state instead of a missing atom.


Deraen:
Please fix the map indent.

PR #514, opened 2020-10-17
Update ManagingState.md

mikew1:
fix indentation of `let`.


mikew1:
Just a simple formatting change I noticed when studying this code.
I'm hoping to try to contribute more to these docs, if that would be of interest.

PR #515, opened 2020-10-17
Update ManagingState.md

mikew1:
Fixes alignment of map entries.

Please let me know, if you'd like help here, whether you'd prefer atomic commits or not for doc fixes?
The alternative would be a single commit containing unrelated doc fixes, so a name like 'various formatting fixes'.


mikew1:
I just noticed (when viewing the file in sublime text) that the source .md file contains tabs, which seems to have something to do with why the result in .html has lost intended formatting (formatting comes out fine in sublime, but is corrupted in the in-browser html). Perhaps the whole file needs a once over to convert tabs to spaces?  Happy to do that if requested.

PR #532, opened 2021-02-28
Support for css modules

bendyorke:
Hello reagent team - 

I recently extended the reagent compiler to include support for a custom tag parser to I could leverage css-modules in my own codebase. I have yet to open source anything, but I included my project code and pretty in depth documentation in a gist here: https://gist.github.com/bendyorke/09328798d468b0d1e7f730bba15e3236

Main idea is that by attaching a css module to the namespace you're in, you can then use namespace qualified keywords to locally scope css class names. i.e `:div.foo` becomes `<div class="foo">` and `::div.foo` becomes something akin to `<div class="__module_foo_abc123">`.

I'm not sure if this is something you're interested in merging in - either the custom tag parsing support or maybe the locally scoped css in general. It's easy enough for me to maintain separately, but css modules are a fantastic feature and namespace qualified keywords really make them incredibly easy to use.

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,3 +1,3 @@
 {:paths ["src" "test" "examples/todomvc/src" "examples/simple/src" "examples/geometry/src" "demo"]
  :deps {org.clojure/clojurescript {:mvn/version "1.10.753"}
-        doo {:mvn/version "0.1.11"}}}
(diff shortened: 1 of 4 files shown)


Deraen:
Merged in #539. I changed to option name to `parse-tag`.

PR #534, opened 2021-03-20
add reagent.dom to ns form for "Essential API" example

oakmac:
It looks like `reagent.dom` is not being included in the `ns` form under the "Essential API" section of the [intro page](http://reagent-project.github.io/).

I think this PR fixes the issue, but I'm not sure what other changes may need to be made in order to push out the fix to the page itself.

### DIFF
--- a/demo/reagentdemo/intro.cljs
+++ b/demo/reagentdemo/intro.cljs
@@ -110,6 +110,9 @@
 (def ns-src (s/syntaxed "(ns example
   (:require [reagent.core :as r]))"))
 
+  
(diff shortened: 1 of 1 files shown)


oakmac:
Thank you for the merge!

FYI - As of this posting (30 Mar 2021) I do not see this change reflected on the public-facing site: https://reagent-project.github.io/

I am not sure what next step is necessary for that? Maybe a "build and publish" command somewhere?

Deraen:
Would be released automatically normally, but the tests have been broken for last few months and I haven't found a fix yet. Updated manually now.

Issue #546, opened 2021-08-28
Add `react` function for working with exteral atom-like types


pepijndevos:
Reagent comes with several atom-like types that implement `IDeref` and `IWatchable`, but to my knowledge does not support implementing your own data sources. In Rum there is a `react` function that basically sets up the watcher and calls deref. Could Reagent support the same? From my limited understanding it'd basically be 

```clojure
(defn react [a]
  (notify-deref-watcher! a)
  @a)
```

This would enable implementing elegant reactive behavior over JS libraries, websockets, localstorage, you name it. Anything that supports taking the value and notifying changes.

Note that `notify-deref-watcher!` itself is a private function, so it's currently not kosher to implement this externally.

Deraen:
Started investigating this in #547

The implementation is indeed quite simple, as Ratom doesn't itself need to do much, just to register itself to reactive context, and the reactive context (reaction) just uses regular IWatchable protocol to listen for changes, and normal atoms implement that.

PR #547, opened 2021-09-03
Test two ways to get regular atoms work in reactive context

Deraen:
Related to #546 

`react` name is already so overloaded that I think we want to avoid that.

Looks like just calling `notify-deref-watcher!` is enough. That function doesn't modify the atom itself, but stores reference to current reactive context to the atom. Reagent will then use `IWatchable` `-add-watch` (and `-remove-watch`) to make reactive context to listen to this atom. Because regular atom implements this protocol same as Ratom.

While it would be simple to just provide single function to use instead of regular `deref` to use with normal atoms, I think better API might be to provide function to wrap atoms, so we can replace deref implementation. This way we could later also modify other methods, if necessary for some reason.

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -628,3 +628,69 @@
                 (flush!)))
         (dispose! res))))
   (ratom-perf))
+
+(deftype WrappedAtom [x]
+  IAtom
+  IReactiveAtom
+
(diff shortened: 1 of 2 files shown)


pepijndevos:
Maybe the watch function should pass the wrapper reference?

PR #555, opened 2021-12-19
Test different ways to declare fn should be used as functional component

Deraen:
I was interested if `:f>` could be replaced with a way to mark the functions using metadata (or something) so that they are used as functional components.

But metadata from defn symbol and defn attr-map are stored to the var metadata, so it is not accessible to Reagent in runtime unless the functional was referred using the var.

Probably it is just best to make functional components the default eventually.


Deraen:
Same idea merged in #638

PR #556, opened 2021-12-19
Fix #548: Support function-component version of reactify-component

### LINKED ISSUE #548: Make possible to reactify-component to return a functional component
I did not found any way to make reactify-component return a functional component.
Even if I use it with a functional compiler like this :
```clojure 
(def functional-compiler (reagent.core/create-compiler {:function-components true}))
(defn reagent-cmp
[props] 
[:> rn/Text "Hello]" )
(def react-cmp (reactify-component reagent-cmp functional-compiler))
```
`react-cmp` will no be a functional component. We can use as-element in the `reagent-cmp` and use the function as a react component but this prevent us to use ratom inside.

Will it be possible to add an option to make reactify-component return a functional component ?

Thanks

Deraen:
TODO: Make it work.

### DIFF
--- a/src/reagent/impl/component.cljs
+++ b/src/reagent/impl/component.cljs
@@ -346,31 +346,51 @@
   (gobj/set c (p/get-id compiler) constructor)
   constructor)
 
-(defn fn-to-class [compiler f]
-  (assert-callable f)
-  (warn-unless (not (and (react-class? f)
-                         (not (reagent-class? f))))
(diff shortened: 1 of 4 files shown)


m-lce:
Thanks for the fix ! Any idea of when this could be merged ?

Deraen:
> Thanks for the fix ! Any idea of when this could be merged ?

It doesn't work yet, and when I worked on this previously I was out of ideas how to make it work really.

PR #559, opened 2021-12-30
Update ManagingState.md

rgkirch:


Deraen:
Thanks!

`auto-run` might also need more docs, at least internally the value can also be a function, but perhaps that is best not exposed to users...

Btw. if you have multiple docs fixes, a single PR would be fine also. Not sure if that is easy to achieve from Github UI though.

PR #560, opened 2021-12-31
Update ManagingState.md

rgkirch:
I actually wanted to log state changes and I was happy to find this example but I think `run!` is a better fit for this purpose. I think it's worth mentioning. IIUC there's nothing to dispose in the `run!` example and no way to stop it. Maybe that's a reason to not bring it up or just something that should get added. Idk.
Thanks for the docs!


Deraen:
> I think it's worth mentioning. IIUC there's nothing to dispose in the run! example and no way to stop it. Maybe that's a reason to not bring it up or just something that should get added.

Yeah. `run!` returns a `Reaction` which is disposable, similar to the `Track` object.

rgkirch:
Thanks for teaching me about that. I'm going to close this for now but I may come back to it in the future after I take time to complete the thought.

Issue #533, opened 2021-03-13
Compile warning: `JSC_UNREACHABLE_CODE`


harold:
First, thanks for reagent, it's nuts :peanuts: - it's the best and funnest way I've found to do frontend, and by not a small margin.

Getting a compile warning, wondering if others are seeing this or if it's an artifact of my (admittedly quite goofy) setup.

I am on 1.0.0:

```
$ lein deps :tree | grep reagent
[reagent "1.0.0"]
```

As part of my production build, this is printed:
```
Mar 13, 2021 2:28:13 PM com.google.javascript.jscomp.LoggerErrorManager println
WARNING: /home/[REDACTED]/cljs-out/prod/reagent/core.js:984: WARNING - [JSC_UNREACHABLE_CODE] unreachable code
return null;
^^^^^^^^^^^^
```

Maybe this helps clean up some code? Regardless, thanks again for your time and consideration.

PR #561, opened 2022-01-10
Add ^ to metadata on reagent.core/render

danielcompton:
Previously builds would get a warning about unreachable code. Adding the ^ marks the deprecated map as metadata and removes the warning.

Fixes https://github.com/reagent-project/reagent/issues/533

```
[:app] Compiling ...
------ WARNING #1 -  -----------------------------------------------------------
 File: ~/.m2/repository/reagent/reagent/1.0.0/reagent-1.0.0.jar!/reagent/core.cljs:370
--------------------------------------------------------------------------------
 367 |                                 tmpl/default-compiler*
 368 |                                 compiler)))
 369 | 
 370 | (defn render
--------------------------------------------------------------------------------
 unreachable code
--------------------------------------------------------------------------------
 371 |   {:deprecated "0.10.0"}
 372 |   [& _]
 373 |   (throw (js/Error. "Reagent.core/render function was moved to reagent.dom namespace in Reagent v1.0."))
 374 |   nil)
--------------------------------------------------------------------------------
```

### DIFF
--- a/src/reagent/core.cljs
+++ b/src/reagent/core.cljs
@@ -368,7 +368,7 @@
                                 compiler)))
 
 (defn render
-  {:deprecated "0.10.0"
-   :superseded-by "reagent.dom/render"}
+  ^{:deprecated "0.10.0"
+    :superseded-by "reagent.dom/render"}
   [& _]
   (throw (js/Error. "Reagent.core/render function was moved to reagent.dom namespace in Reagent v1.0.")))


danielcompton:
Hmm, it might be that the previous version was correct and this warning is incorrect?

Deraen:
The warning is about `nil` return value.

Defn takes metadata map as second parameter after the name. If adding metadata to the var using `^` the map needs to be before the fn name, so yes, the metadata is already correct.

Deraen:
https://github.com/reagent-project/reagent/commit/d2894ec7d6112b1ea4bdd7102dde2098c5939390#diff-cc4bc26db849be7e3561fa1b7c0b7314120f689bfaea74b712aa8748322dcbd7

The unncessary return value should be already fixed in v1.1.0: https://github.com/reagent-project/reagent/commit/d2894ec7d6112b1ea4bdd7102dde2098c5939390#diff-cc4bc26db849be7e3561fa1b7c0b7314120f689bfaea74b712aa8748322dcbd7

danielcompton:
Oh, sorry!

PR #562, opened 2022-01-28
Add meta to RCursor.

rgkirch:
e.g. to support posh!

```clojure
(defonce app-db (r/atom (or (unsave-app-db) {})))

(defonce conn (with-meta (r/cursor app-db [:ds-db]) {:listeners (atom {})}))
(pr/posh! conn)
```


rgkirch:
🎉

PR #567, opened 2022-06-06
fix left-quote in intro

shaunlebron:
small fix:

before:
CleanShot 2022-06-06 at 14 28 19@2x

after:
CleanShot 2022-06-06 at 14 28 35@2x


shaunlebron:
Oops, there’s more to this than I thought, but the English row shows it: https://en.wikipedia.org/wiki/Quotation_mark#Summary_table

PR #577, opened 2022-08-26
Refactor Reaction type to clarify when things “change”

shaunlebron:
After having trouble understanding all the branches inside the Reaction Deref function, I pulled the `notify-deref-watcher!` before the dirty check, so we can combine both dirty checks.  Please verify that this is equivalent and easier to follow!

### DIFF
--- a/src/reagent/ratom.cljs
+++ b/src/reagent/ratom.cljs
@@ -391,14 +391,13 @@
       (-deref this)))
 
   (_handle-change [this sender oldval newval]
-    (when-not (or (identical? oldval newval)
-                  dirty?)
-      (if (nil? auto-run)
-        (do
-          (set! dirty? true)
-          (rea-enqueue this))
-        (if (true? auto-run)
-          (._run this false)
+    (when-not dirty?
+      ;; Use `identical?` to determine equality from input changes.
+      (when-not (identical? oldval newval)
+        (case auto-run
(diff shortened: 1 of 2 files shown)


shaunlebron:
Also, @holmsand, could you explain the purpose of this branch?

```clj
(if (and non-reactive (nil? auto-run))  ;; <-- what exactly does this mean?
  …)
```

src/reagent/ratom.cljs
```diff
@@ -456,19 +456,17 @@
...
-          (let [oldstate state]
-            (set! state (f))
-            (when-not (or (nil? watches) (= oldstate state))
-              (notify-w this oldstate state))))
-        (do
-          (notify-deref-watcher! this)
-          (when dirty?
-            (._run this false)))))
+    (if (reactive?)
+      (notify-deref-watcher! this)
```
Deraen, line 460: Previously the branch where `notify-deref-watcher!` is called was also run if `(nil? auto-run)` was falsey.

```
(and non-reactive (nil? auto-run))
(or (not non-reactive) (not (nil? auto-run)))
(or reactive (some? auto-run))
```

Tests seem to pass, so not sure if this is important. Or this case may be missing a test.

shaunlebron:
I’ll start a new PR to clarify these extra changes.

PR #593, opened 2023-06-28
Emotion example

cj-price:
An example of a reagent compiler that implements [emotion](https://emotion.sh/docs/introduction).


Deraen:
Hi,

I'm not sure about merging this. While the example is good, I fear the complexity of what is required for Emotion integration is too much to keep up-to-date in the Reagent examples folder.

The core ns with the use examples is fine, but the compiler and syntax namespaces seem more suited to a real library or a separate example project.

PR #594, opened 2023-08-09
Introduce pure clojure wrappers for React Context

ingesolvoll:
This PR suggests a pure Clojure API for React Context. It's a very simple wrapper around `Provider` and `Consumer`. It does not convert the context value in any way.

If the Reagent maintainers decide that this abstraction is useful and within relevant scope, I will complete this PR with tests, docs, and any other things missing in order to be considered for merge.

A short code example:

```clojure
(require '[reagent.context :as r.context])

(def my-context (r.context/make-context "my-context" "default"))

[r.context/provider {:value {:foo :bar} ;; The value here can be anything, does not need to be a map
                     :context my-context}
    [r.context/consumer {:context my-context}
     (fn [{:keys [foo]}]
       [:div ":foo is a keyword: " (pr-str foo)])]

    ;; The `with-context` macro cuts away some boilerplate from the above
    (r.context/with-context [{:keys [foo]} my-context]
      [:div "From the macro: " (pr-str foo)])]
```

The code in this PR was written by the developers at [Whimsical](https://whimsical.com)

### DIFF
--- a/src/reagent/context.cljs
+++ b/src/reagent/context.cljs
@@ -1,0 +1,42 @@
+(ns reagent.context
+  (:require-macros [reagent.core]
+                   [reagent.context])
+  (:require [react :as react]
+            [reagent.core :as r]))
+
+(defn make-context
+  "Creates a context with the given name and optional default value.
+   The default value will be used when trying to retrieve the context value with no provider present.
+   Attempting to retrieve a context value that has no provider, and no default value will result in a crash."
+  ([name]
+   (let [context (react/createContext)]
+     (set! (.-displayName context) name)
+     context))
+  ([name default-value]
+   (let [context (react/createContext default-value)]
+     (set! (.-displayName context) name)
+     context)))
+
+(defn provider
+  "Provides the value for the given context to descendant components."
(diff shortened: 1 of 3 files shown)


src/reagent/context.cljs
```diff
@@ -0,0 +1,42 @@
...
+  ;; remount every time a new prop comes into the parent.
+  (r/with-let [wrapper-comp
+               ;; Passes through context to component using meta data. See:
+               ;; https://github.com/reagent-project/reagent/blob/ce80585e9aebe0a6df09bda1530773aa512f6103/doc/ReactFeatures.md#context
+               ^{:context-type context}
+               (fn [render-f]
+                 (let [value (.-context (r/current-component))]
+                   (when (undefined? value)
+                     (throw (js/Error. (str "Missing provider for " (.-displayName context)))))
+                   (render-f value)))]
```
lilactown, line 41: one of the problems with using the `Consumer` component is that you couldn't deref ratoms safely inside the render fn; they wouldn't setup a listener because they wouldn't have the reactive context. Does this fix that?

PR #598, opened 2023-12-01
Add defc macro to create fn components with Fast Refresh support

Deraen:
Investigating if React Refresh support is possible in some way. It will require using a macro, because attaching necessary data to the function (component) just isn't possible when components are created when rendering Hiccup forms (what Reagent normally does.)

- Target is to get React Refresh working
- Working React Refresh will allow better dev experience on a project transitioning from Reagent to UIx/Helix
- As a side-effect this will create function components which work without `:f>`
- Macro to handle Hiccup or properties transform is not a target (one could probably wrap the macro with Sablono to do that)

TODO:
- Looks like form-2 components don't work
- Form-3 not tested, not sure if makes even sense


doc/ReactRefresh.md
```diff
@@ -0,0 +1,8 @@
+# React Refresh
+
+- `reagent.dev` ns must be required before anything that loads `react-dom`
+- Don't call `r.dom/render` after reload (e.g. shadow-cljs hook)
+- Call `reagent.dev/refresh!` instead
+- Only components defined using `r/defc` will refresh
+- Reagent doesn't try to create Hook signatures for components,
+  so hook state is reset for updated components.
```
roman01la, line 8: I’m curious, if local state will be reset, what’s the motivation to enable react refresh in Reagent?

Deraen: To help projects using both Reagent and UIx/Helix, so they can enable it and get benefit on React Refresh for UIx/Helix components.

PR #607, opened 2024-06-24
Add documentation/example for SSR

MonsieurV:
A simple code example of server-side rendering with `reagent.dom.server/render-to-string`

Related #302


MonsieurV:
Hi @Deraen,

I've seen I've added `.clj-kondo` to `.gitignore`. I'm not sure if that was right, maybe `.clj-kondo/.cache` would be better.

Issue #504, opened 2020-06-29
Investigate removing rAF scheduling


Deraen:
https://gist.github.com/roman01la/8b65652af9d086e70aabfdf08ba6129c#removed-raf-based-scheduling-mechanism

Would remove need for input workarounds and make interop easier.

Deraen:
This didn't seem possible: https://github.com/reagent-project/reagent/pull/612

PR #612, opened 2024-11-15
Testing out disabling Reagent ratom/rendering queue

Deraen:
#504 etc.

Currently test suite is getting stuck due to test using `u/with-mounted-component-async` and `u/run-fns-after-render` not finishing at all.

TODO:
- [x] ~Figure out how to call after-render callbacks after React finishes the render~
- [x] New render and act utils to run checks after component is rendered
- [ ] Check and consider if we can just completely drop `create-class` and the class-based component implementation
    - Probably not a required change, but would make reagent code cleaner
- [ ] drop reagent.core/after-render and possibly other queue functions
    - There is no simple way to hook into React rendering queue to keep this working as previously.
      Likely this isn't used that often, so users would just replace these places with hooks manually?

### DIFF
--- a/test/reagenttest/testreagent.cljs
+++ b/test/reagenttest/testreagent.cljs
@@ -926,6 +926,7 @@
 (defn foo []
   [:div])
 
+#_
 (u/deftest ^:dom test-err-messages
   (when (dev?)
     (is (thrown-with-msg?
@@ -1065,6 +1066,8 @@
   (is (= "<p>#object[reagent.ratom.RAtom {:val 1}]</p>"
          (as-string [:p (r/atom 1)]))))
 
+;; FIXME: r/after-render won't work
+#_
 (u/deftest ^:dom test-after-render
   (let [spy (atom 0)
         val (atom 0)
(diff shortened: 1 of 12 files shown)


test/reagenttest/testreagent.cljs
```diff
@@ -61,71 +66,74 @@
...
+              (reset! val 1)
+              (reset! val 2)
+              (reset! val 1))
+
+            ;; (r/flush)
+            (is (= "val 1 1 0" (.-innerText div)))
+            (is (= 2 @ran) "ran once more")
+            ;; NOTE: OKAY Here is a problem:
+            ;; reactions are now being run for each input ratom change because we don't have the queue anymore!
+            (is (= 2 @reaction-ran))
```
Deraen, line 123: As I was thinking, while components aren't rendered for each Ratom change due to the React queue, here the    `reaction` body is running each time input Ratoms change, which is potentially a (big) problem for re-frame.

Consider cases where text input value is used as a reaction (or subscription) input.

Deraen:
This version likely wont work.

Issue #610, opened 2024-10-08
Export clj-kondo config


Deraen:
https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting-and-importing-configuration

Reagent has at least the `with-let` macro that would benefit from clj-kondo configuration. At the simplest, it can be a lint-as `let` option, but we could also add a hook that allows the `finally` form.

PR #614, opened 2025-01-15
Add a basic clj-kondo config for library consumers

souenzzo:
Address part of what is described in

- https://github.com/reagent-project/reagent/issues/610

I think that this small contribution

- makes it easier for "newbies" to use reagent in a kondo-powered IDE (clojure-lsp is powered by kondo)
- facilitates and avoids explicit kondo configuration for simple use-cases

I plan to do a full implementation (via hooks, supporting finally) soon.


Deraen:
Thanks!

PR #622, opened 2025-02-10
Update README.md

Carl-Porter:
Specifically reagent.dom/render function throws an error on newer versions of react, and likely for other functions as well. Updating readme to specify that you need to npm install specific versions of react and react-dom for shadow-cljs builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ To use Reagent in an existing project you add this to your dependencies in `proj
 And provide React using either npm (when using e.g. [Shadow-cljs](https://shadow-cljs.github.io/docs/UsersGuide.html))
 
 ```
-npm i react react-dom@17.0.2 react@17.0.2
+npm i react react-dom@18.3.1 react@18.3.1
 ```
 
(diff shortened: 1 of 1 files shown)


Deraen:
We should likely refer to React 18.3.1.

Carl-Porter:
Ah that works as well. I saw [this](https://github.com/reagent-project/reagent/discussions/611) discussion and assumed React 18 was not ideal. I updated the change to specify 18.3.1 and specifically call out that React 19 is the one that breaks reagent-dom/render

Carl-Porter:
Actually looking at it again, it runs with 18.3.1 but it's automatically outputting this error message. Is there a good reason to run 18.3.1 over 17.0.2?
 
 

> ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17.

Deraen:
Using React 18 in the compatibility mode is fine. It is just a warning. 

I'm not sure, but I guess some of the React 18 features are available in the compatibility mode, and some would require the new createRoot use, with the concurrent mode. Some libs might also require React 18.

PR #624, opened 2025-03-02
Fix geometry example build

BenSturmfels:
Updated the Shadow CLJS :init-fn to refer to the `run` function.

Before this change, `npm shadow-cljs watch :app` failed with:

```
[:app] Build failure:
Invalid configuration
-- Spec failed --------------------

  {:target ...,
   :output-dir ...,
   :asset-path ...,
   :devtools ...,
   :modules {:main {:entries ..., :init-fn geometry.core}},
                                           ^^^^^^^^^^^^^
   :build-id ...}

should satisfy

  unquoted-qualified-symbol?
```


BenSturmfels:
I meant to say, thanks so much for Reagent, I really appreciate your work!

Issue #597, opened 2023-11-25
Controlled input loses cursor under ShadowRoot in React 18


kimo-k:
This [controlled-input cursor bug](https://github.com/reagent-project/reagent/blob/master/doc/ControlledInputs.md) appears when I try to use the new `react.dom.client` API to render an `:input` component within a shadow root. The text cursor skips to the end every time I insert a character.

The same component works fine with the `react.dom` API.

Is this expected, and is there anything I can do? I'm reticent to go with the [react-native workaround](https://github.com/reagent-project/reagent/blob/master/doc/ControlledInputs.md#react-native), since my codebase is pretty large, with everything inside a shadow root (i.e. re-frame-10x).

Minimal repro: https://github.com/reagent-project/reagent/compare/master...kimo-k:reagent:shadow-root-controlled-input-bug

jaidetree:
Not sure if this fixes the issue but that code doesn't seem quite right. 

It's creating a new input element every render cycle instead of updating an existing one. You could try adding a `:key` property to see if that improves the tracking. When I've ran into similar issues, it was usually because it was deleting the previous dom element and replacing it with a new one instead of updating it.

Lastly it seems like it's better to use `with-let` to initialize your atom in the `repro` component definition so you don't have to return a function. That way the atom is created only once in the component's lifecycle.

```clojure
(defn repro []
  (with-let [text (r/atom "")]
    [:input {:on-change #(reset! text (.. % -target -value))
             :value @text}]))
```

Not sure that will entirely fix it, but it's a place to start.

kimo-k:
Hey @jaidetree, thanks for taking a look. Were you able to run the repro without bugs?

I think your hypothesis might be incorrect, though. As far as I know, `repro` is an idiomatic form-2 component, `:key` would have no effect outside a seq, and using `with-let` will macroexpand to the same code.

Deraen:
@kimo-k Just note about `:key`. It isn't specific to sequences at all, any React component can have a key and React will use it to detect if the component identity changes. If identity changes, the previous component is unmounted and a new one created. This is different to just updating properties for same component instance. This can be useful outside of seqs when you need to reinitialize component local state (ratoms or hook state).

`with-let` implementation also does other stuff than just macroexpand into a basic form-2 component. Though it does indeed expand into form-2 component.

Reagents controlled input cursor hacks are a huge mess and our async rendering batching for ratom changes is really problematic with React 18 update batching so there is probably no easy fix for this.

[...]

PR #627, opened 2025-04-22
Fix cursor jumping for inputs inside shadow dom

oxalorg:
Look at #619 and #597 

## Problem

Reagent already has logic to not lose cursor position for controlled text inputs, but it only works when the current node in focus is also the active element.

When you're inside a shadow dom, the `document.activeElement` is set to shadow root even when the active element is an input inside the shadow dom (which makes sense given that shadow dom is an encapsulated piece whose dom access should not be apparent to the outside world immediately)

Which means that when you're trying to type inside a controlled input element, the focus keeps resetting and the cursor keeps jumping.

## Technical description

So what we know is that if an input element is inside a shadow dom, the `activeElement` points to the shadow root instead of the input element. This messes up the cursor perservation logic. 

So the solution should be simple, we recurse down to find the active element inside the shadow dom. And hence the saving grace here is that we can do exactly that by querying for the `.activeElement` on the shadow-root. This PR does that.

### DIFF
--- a/src/reagent/impl/input.cljs
+++ b/src/reagent/impl/input.cljs
@@ -17,9 +17,25 @@
 
 (def ^:dynamic *force-set-dom-value* false)
 
+(defn get-true-active-element []
+  ;; For controlled inputs `input-node-set-value` tries to do some heurestics to
+  ;; preserve the cursor position. But before doing so, it tries to find the active
+  ;; element in the document.
+  ;;
+  ;; Unfortunately, if an input element is inside a shadow dom, the
+  ;; activeElement points to the shadow root instead of the input element. This
+  ;; messes up the cursor perservation logic.
+  ;;
+  ;; But the saving grace here is that active input element can still be found
+  ;; by querying for the `.activeElement`` on the shadow-root.
+  (let [active-element (.-activeElement js/document)]
+    (if-let [shadow-root (.-shadowRoot active-element)]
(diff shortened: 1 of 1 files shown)


Deraen:
Looks good on a quick glance. I'll merge this when I next have time to work on Reagent.

One thing to consider: is it possible there are multiple nsted shadowRoots? If it is possible, not sure how likely such case is.

oxalorg:
@Deraen I'll create a repository reproducing this issue and share here either tomorrow or over the weekend! ^_^

oxalorg:
@Deraen Here is a minimal reproduction of the issue. There is also a branch which applies this fix (using [GaiwanTeam](https://github.com/gaiwanTeam/reagent)) so you can test both the behaviours!

https://github.com/oxalorg/reagent-cursor-jump-repro 

LMK if I should add something more specific to this reproduction? Thanks and apologies for the delayed response!

Deraen:
@oxalorg, Thanks! The example looks good.

I will consider if/how I can incorporate this in Reagent test suite, or if I just have to keep it as an example project and run manual testing.

I will be on vacation for most of August so it could be some time before I get around to merging these, but with the example case, I think I now have everything I need.

Deraen:
Adding a setTimeout 0 to the state update ensures the cursor is broken like expected. This is similar to what triggering a re-frame event and using a subscription to get the value would do.

PR #632, opened 2025-07-11
Rework run-in-reaction to avoid the magic to store reaction instace t…

Deraen:
…o the obj


Deraen:
NOTE: uix (use-subscribe) and maybe some others use the old run-in-reaction.

And are there cases where someone is trying to access cljsRatom property? Hopefully not.

Issue #634, opened 2025-07-22
Functional components forget their watched atoms in strict mode


matej-ibis-ai:
Running with React strict mode (enabled at the top-level component), React effects are cycled one extra time, which causes functional components created using `reagent.impl.component.functional-render` to dispose of their watched values, and not watch them anymore.

The effect defined in `functional-render` should provide a symmetrical action to `ratom/dispose!` in its first part, in the `mount` function.

In my preliminary tests, I have extended the `Reaction` type for the following two methods:
```clojure
(_snapshot [_this]
  [state watching auto-run])
                                                      
(_restore [this [s w a]]
  (set! watching w)
  (set! state s)
  (set! auto-run a)
  (set! dirty? true)
  (doseq [watcher (set w)]
    (-add-watch watcher this handle-reaction-change)))
```
and used them to save the related values of the Reaction in a new Javascript field between the cleanup and second run of the effect, like this:
```clojure
(react/useEffect
 (fn mount []
   (when-some [reagent-state (.-current state-ref)]
     (when-some [r-argv (gobj/get reagent-state "ratomSnapshot" nil)]
       (let [rat (gobj/get reagent-state "cljsRatom")]
         (._restore rat r-argv)
         (gobj/remove reagent-state "ratomSnapshot"))))
   (fn unmount []
     (when-some [rat ^ratom/Reaction (gobj/get reagent-state "cljsRatom")]
       (gobj/set reagent-state "ratomSnapshot" (._snapshot rat)))
     (some-> (gobj/get reagent-state "cljsRatom") ratom/dispose!)
     ))
 ;; Ignore props - only run effect once on mount and unmount
 ;; (which means always twice under the React strict mode).
 #js [])
```

That solved the issue I have described here, although I still see failures in expanding seqs rendered by these functional components -- but I will be filing that as a separate issue when I gain more understanding of what's going wrong there.

PR #635, opened 2025-07-29
Add the :strict? option to with-render*

matej-ibis-ai:
Setting the `:strict?` option causes the react/StrictMode component to be inserted at the top level, wrapping Reagent root.

This is a prerequisite for fixing #634, which I am going to address next.

### DIFF
--- a/src/reagent/dom/client.cljs
+++ b/src/reagent/dom/client.cljs
@@ -51,13 +51,19 @@
   "Render the given Reagent element (i.e. Hiccup data)
   into a given React root."
   ([^js root el]
-   (render root el tmpl/*current-default-compiler*))
+   (render root el nil))
   ([^js root el compiler]
+   (render root el compiler false))
+  ([^js root el compiler strict-mode?]
    (set! batch/react-flush react-dom/flushSync)
-   (let [comp (fn [] (p/as-element compiler el))
-         js-props #js {}]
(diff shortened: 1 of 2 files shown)


src/reagent/dom/client.cljs
```diff
@@ -51,13 +51,19 @@
...
+   (render root el compiler false))
+  ([^js root el compiler strict-mode?]
    (set! batch/react-flush react-dom/flushSync)
-   (let [comp (fn [] (p/as-element compiler el))
-         js-props #js {}]
-     (set! (.-comp js-props) comp)
-     (.render root (react/createElement reagent-root js-props)))))
+   (let [eff-compiler (or compiler tmpl/*current-default-compiler*)
+         comp (fn [] (p/as-element eff-compiler el))
+         js-props #js {:comp comp}
```
Deraen, line 61: This doesn't work with advanced optimization. JS object literal property name can't be renamed/manged on the optimization step.

The `reagent-root` function is accessing `.-comp` property, which will be renamed.

matej-ibis-ai: Oh... Indeed, I haven't tried to grasp the differences in ways to access Javascript object properties from Clojurescript yet, so it doesn't surprise me I make errors in this. But I see you merged my commit -- thanks! -- so I assume you are going to deal with this bug in some followup commits.

PR #636, opened 2025-07-29
Make functional components work in strict mode

matej-ibis-ai:
Building on #635, this PR
* adds tests showing how functional components break under the React Strict Mode; and
* proposes a solution to the issue, updating the effect in `functional-render` to make it idempotent, to solve the related issue, #634.

I have retained commits with some other approaches I tried but which didn't work. The current latest commit contains only the solution which seems to work well.

### DIFF
--- a/src/reagent/impl/component.cljs
+++ b/src/reagent/impl/component.cljs
@@ -429,23 +429,11 @@
 
       (react/useEffect
        (fn mount []
-         (when-not (gobj/containsKey reagent-state "cljsRatom")
-           (ratom/run-in-reaction
-            ; `run` function, its derefs get captured and it's the one executed
-            ; by the `Reaction._run` method.
-            #(functional-do-render compiler reagent-state)
-            ; Container for the reaction.
-            reagent-state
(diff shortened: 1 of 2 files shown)


Deraen:
I started investigating the issue.

"Proper" way for a React Effect to work, would be that the entry call would also initialize the state like regular. We can't do that. The ratom is initialized by render call, and we can't do any tricks to move that render call to happen inside useEffect, because render call might contain other Hook uses, and they can't be called inside the Effect hook.

Another idea I considered to avoid potential problems with garbage collection, was to store the snapshot values into the component closure, like this:

```
          rat ^ratom/Reaction (gobj/get reagent-state "cljsRatom")
          rat-snapshot #js {:val nil}]
```

because component render isn't called again between entry->dispose->entry, the same object would be visible for second effect entry and this value could be used to restore the ratoms. In this case the snapshot value is disposed when render is called again, which I think would ensure the references to old snapshots aren't kept for as long? Maybe. But this approach gets inconvenient because there could be many disposes ratoms, when dispose recursively disposes any watched atoms.

